PR #15, opened 2022-01-25
Update context

JSAssassin:
Updated JSON-LD Context based on the spec and replaced ZLIB  with GZIP.

@msporny Also, I could be wrong but shouldn't the type here be https://github.com/w3c-ccg/vc-status-list-2021/blob/main/index.html#L281 `RevocationList2021Status` and this one https://github.com/w3c-ccg/vc-status-list-2021/blob/main/index.html#L359 be `SuspensionList2021`?

### DIFF
--- a/contexts/v1.jsonld
+++ b/contexts/v1.jsonld
@@ -4,7 +4,7 @@
 
     "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-status-list-2021/#StatusList2021Credential",
+        "https://w3id.org/vc-status-list-2021#StatusList2021Credential",
       "@context": {
         "@protected": true,
 
--- a/contexts/v1.jsonld
+++ b/contexts/v1.jsonld
@@ -4,7 +4,7 @@
 
     "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-status-list-2021#StatusList2021Credential",
+        "https://w3id.org/vc/status-list#StatusList2021Credential",
       "@context": {
         "@protected": true,
 
@@ -18,20 +18,20 @@
 
     "RevocationList2021": {
       "@id":
-        "https://w3id.org/vc-status-list-2021#RevocationList2021",
+        "https://w3id.org/vc/status-list#RevocationList2021",
       "@context": {
         "@protected": true,
 
         "id": "@id",
         "type": "@type",
 
-        "encodedList": "https://w3id.org/vc-status-list-2021#encodedList"
+        "encodedList": "https://w3id.org/vc/status-list#encodedList"
       }
     },
 
     "RevocationList2021Status": {
       "@id":
-        "https://w3id.org/vc-status-list-2021#RevocationList2021Status",
+        "https://w3id.org/vc/status-list#RevocationList2021Status",
       "@context": {
         "@protected": true,
 
@@ -40,17 +40,17 @@
 
         "statusListCredential": {
           "@id":
-            "https://w3id.org/vc-status-list-2021#statusListCredential",
+            "https://w3id.org/vc/status-list#statusListCredential",
           "@type": "@id"
         },
         "statusListIndex":
-          "https://w3id.org/vc-status-list-2021#statusListIndex"
+          "https://w3id.org/vc/status-list#statusListIndex"
       }
     },
 
     "SuspensionList2021": {
       "@id":
-        "https://w3id.org/vc-status-list-2021#SuspensionList2021",
+        "https://w3id.org/vc/status-list#SuspensionList2021",
       "@context": {
         "@protected": true,
 
@@ -59,11 +59,11 @@
 
         "statusListCredential": {
           "@id":
-            "https://w3id.org/vc-status-list-2021#statusListCredential",
+            "https://w3id.org/vc/status-list#statusListCredential",
           "@type": "@id"
         },
         "statusListIndex":
-          "https://w3id.org/vc-status-list-2021#statusListIndex"
+          "https://w3id.org/vc/status-list#statusListIndex"
       }
     }
   }
--- a/contexts/v1.jsonld
+++ b/contexts/v1.jsonld
@@ -2,9 +2,9 @@
   "@context": {
     "@protected": true,
 
-    "RevocationList2020Credential": {
+    "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-revocation-list-2020#RevocationList2020Credential",
+        "https://w3id.org/vc-status-list-2021/#StatusList2021Credential",
       "@context": {
         "@protected": true,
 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -232,7 +232,7 @@ <h2>Core Concept</h2>
              list of boxes at the top of the image with two of them in
              red depicting revoked credentials. Text beside the boxes to the
              right reads 16 kilobytes. An depiction of the boxes being
-             ZLIB compressed into a cylinder on the bottom of the page shows
+             GZIP compressed into a cylinder on the bottom of the page shows
```
OR13, line 235: I have been using https://www.npmjs.com/package/pako

Can we add an informative reference to an implementation you have in mind?

I have seen some really nasty issues with non-deterministic / non-idempotent compression before, and would prefer to avoid accidentally taking a dependency like that.

dmitrizagidulin (approved):

contexts/v1.jsonld
```diff
@@ -2,9 +2,9 @@
   "@context": {
     "@protected": true,
 
-    "RevocationList2020Credential": {
+    "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-revocation-list-2020#RevocationList2020Credential",
+        "https://w3id.org/vc-status-list-2021/#StatusList2021Credential",
```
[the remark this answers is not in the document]
OR13: ahh yes... this .... pretty much everywhere...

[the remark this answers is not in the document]
OR13: do it now, before it breaks signatures.

index.html
JSAssassin: @OR13  DB is using gzip/ungzip from pako for the vc-revocation-list 2020 implementation. 
Here's a gist of the example encodedList value for that
https://gist.github.com/JSAssassin/86f92c6cbdd712e6a6db5f78fd31ba6d. 
Is there anything else you'd like to see?

contexts/v1.jsonld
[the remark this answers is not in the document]
JSAssassin: @OR13 @dmitrizagidulin @dlongley thoughts on this? ^ Do we want to make the change this time?

index.html
OR13: can you link me to the npm package you are using? / unit test for it?

contexts/v1.jsonld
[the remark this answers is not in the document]
OR13: I would tackle it in a subsequent PR, the proposal is blocked by updates to w3id.org.... assume that won't happen until it does.

index.html
JSAssassin: @OR13 this is the test suite I am using https://github.com/digitalbazaar/status-list-2021-test-suite/tree/initial, even though the name of the test suite is `status-list-2021`, since 2021 stuff has not been implemented yet it's using endpoints that uses the vc-revocation-list 2020 implementation.

contexts/v1.jsonld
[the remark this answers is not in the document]
dlongley: @JSAssassin, yes, I agree with @msporny, we don't want a date in the *base URL* for the vocab.

contexts/v1.jsonld
```diff
@@ -2,9 +2,9 @@
   "@context": {
     "@protected": true,
 
-    "RevocationList2020Credential": {
+    "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-revocation-list-2020#RevocationList2020Credential",
+        "https://w3id.org/vc/status-list#StatusList2021Credential",
```
OR13, line 7: this link is dead.

https://w3id.org/vc/status-list#StatusList2021Credential

OR13 (changes requested):
No context changes that include dead links.

contexts/v1.jsonld
JSAssassin: The link should be working now.  w3id.org has been updated.

contexts/v1.jsonld
```diff
@@ -2,9 +2,9 @@
   "@context": {
     "@protected": true,
 
-    "RevocationList2020Credential": {
+    "StatusList2021Credential": {
       "@id":
-        "https://w3id.org/vc-revocation-list-2020#RevocationList2020Credential",
+        "https://w3id.org/vc/status-list#statuslist2021credential",
```
msporny, line 7: Capitalization incorrect here.

contexts/v1.jsonld
```diff
@@ -16,35 +16,54 @@
...
-        "encodedList": "https://w3id.org/vc-revocation-list-2020#encodedList"
+        "encodedList": "https://w3id.org/vc/status-list#encodedList"
       }
     },
 
-    "RevocationList2020Status": {
+    "RevocationList2021Status": {
       "@id":
-        "https://w3id.org/vc-revocation-list-2020#RevocationList2020Status",
+        "https://w3id.org/vc/status-list#revocationlist2021status",
```
msporny, line 34: Capitalization incorrect.

contexts/v1.jsonld
```diff
@@ -16,35 +16,54 @@
...
         "@protected": true,
 
         "id": "@id",
         "type": "@type",
 
-        "revocationListCredential": {
+        "statusListCredential": {
           "@id":
-            "https://w3id.org/vc-revocation-list-2020#revocationListCredential",
+            "https://w3id.org/vc/status-list#statusListCredential-1",
```
msporny, line 43: Capitalization incorrect.

contexts/v1.jsonld
```diff
@@ -16,35 +16,54 @@
...
-        "revocationListIndex":
-          "https://w3id.org/vc-revocation-list-2020#revocationListIndex"
+        "statusListIndex":
+          "https://w3id.org/vc/status-list#statusListIndex-1"
+      }
+    },
+
+    "SuspensionList2021Status": {
+      "@id":
+        "https://w3id.org/vc/status-list#suspensionlist2021status",
```
msporny, line 53: Capitalization incorrect.

contexts/v1.jsonld
```diff
@@ -16,35 +16,54 @@
...
+        "statusListCredential": {
           "@id":
-            "https://w3id.org/vc-revocation-list-2020#revocationListCredential",
+            "https://w3id.org/vc/status-list#statusListCredential-1",
           "@type": "@id"
         },
-        "revocationListIndex":
-          "https://w3id.org/vc-revocation-list-2020#revocationListIndex"
+        "statusListIndex":
+          "https://w3id.org/vc/status-list#statusListIndex-1"
```
msporny, line 47: Term and capitalization incorrect.

msporny (changes requested):
The vocabulary terms need to point to a vocabulary document (which doesn't need to exist yet), not the specification.

contexts/v1.jsonld
JSAssassin: Fixed.

msporny:
@OR13 wrote:
> I require an updated test vector / example encoded value to validate the GZIP implementation, ideally with a proof that it is indeed deterministic / idempotent.

Yes, agreed, raised as issue https://github.com/w3c-ccg/vc-status-list-2021/issues/16.

msporny:
> No context changes that include dead links.

Tracked via issue https://github.com/w3c-ccg/vc-status-list-2021/issues/17.

Issue #3, opened 2021-04-06
Context about how this differs from 2020-rl


swcurran:
It would be very helpful if the specification referenced the vc-revocation-list-2020 and defined the differences between the two. I've been trying to scan to figure out the differences, and from what I can see a couple of mentions (and an example) of "SuspensionList" is the only change.  Is that correct?

A related question is whether the VC for the list have a type that MUST be "RevocationList2021" as stated, or would it also allow "SuspensionList2021" to align with the addition of allowing SuspensionList in the VC given to the holder?

If that is the case, I would add my name to the question in #2 asking why not add a "call it whatever you want" `listtype` field so you aren't limited to revoke and suspend.

Thanks!

csuwildcat:
This is the direction I would hope to go - just allow a field like `topic` where the status topic can be whatever the list implementer wants (possibly with a registry of known values and their meanings).

csuwildcat:
@msporny as we begin implementing this in our production stuff, we would really like to see the concept of topic/list-type. This is important because what we really want is to have all the lists be `StatusList2021`, but have that status topic be different depending on the need/goal. The concept of a topic/list-type would make that easier, imo.

kezike:
Agreed with the points here, especially considering that there is seemingly no material difference between `SuspensionList2021` and `RevocationList2021` in the current version of this spec.

msporny:
> This is the direction I would hope to go - just allow a field like `topic` where the status topic can be whatever the list implementer wants (possibly with a registry of known values and their meanings).

Feels reasonable, especially since we don't have a test suite for interop yet... we'd want to get this locked in before one is created. I'm trying to think through how the change might blow up in our faces... I haven't had a chance to think deeply about that so would prefer some implementer feedback before we commit to the new design path.

The old design clearly specifies the type in both the VC that points to the status list, and the status list itself (so there can be no confusion on what the list is for). If we were to do this, there would be an extra type check required to make sure the topics line up... if we don't do that, there could be an attack where you swap a suspension list for a revocation list (maybe?!)... I don't know, but someone needs to think about the attacks around this deeply before we implement it. Has anyone thought through how one would attack a topic mismatch?

sbutterfield:
@msporny 
Currently going through an implementation cycle of this myself, I have some thoughts:

[...]

Issue #2, opened 2021-02-27
Add status list type field


csuwildcat:
Would it be possible to add a field to note the type of status a status list is for? Like for example: `listType: "revocation"`, where you could put 'suspended', 'pending', or any other status list intent someone wanted?

bumblefudge:
On today's VC-EDU call, some other list types were discussed: "endorsed" or "certified" or "notarized" might be other lists worth exploring for other use cases...

bumblefudge:
[This issue](https://github.com/w3c/vc-data-model/issues/820) over on the VC data model repo also wonders if "refreshed" is a status worth defining (whether in a list type or otherwise)

msporny:
This has been implemented in the latest specification:

```
    "statusPurpose": "revocation | suspension | anyOtherTextStringYouCanImagine"
```

More here: https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021entry

PR #24, opened 2022-03-12
Rework the data model to support `statusPurpose` field.

msporny:
This PR reworks the data model to address issue #2 and #3 by doing the following:

* The `type` field now includes a year... `StatusList2021`.
* The `statusPurpose` field has been added.
* The `statusPurpose` field contains a text string instead of a JSON-LD typed value.
* The `RevocationListX` and `SuspensionListX` types have been removed.
* Two text string status purposes "revocation" and "suspension" have been added.
* The JSON-LD Context and redirects have been fixed/updated.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-status-list-2021/pull/24.html" title="Last updated on Apr 4, 2022, 9:30 PM UTC (01a681d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/vc-status-list-2021/24/abd4f1d...01a681d.html" title="Last updated on Apr 4, 2022, 9:30 PM UTC (01a681d)">Diff</a>

### DIFF
--- a/contexts/v1.jsonld
+++ b/contexts/v1.jsonld
@@ -2,7 +2,7 @@
   "@context": {
     "@protected": true,
 
-    "StatusListCredential2021": {
+    "StatusList2021Credential": {
       "@id":
         "https://w3id.org/vc/status-list#StatusList2021Credential",
       "@context": {
@@ -31,9 +31,9 @@
       }
     },
 
-    "StatusListEntry2021": {
+    "StatusList2021Entry": {
       "@id":
-        "https://w3id.org/vc/status-list#StatusListEntry2021",
+        "https://w3id.org/vc/status-list#StatusList2021Entry",
       "@context": {
         "@protected": true,
 
--- a/contexts/v1.jsonld
+++ b/contexts/v1.jsonld
@@ -2,7 +2,7 @@
   "@context": {
     "@protected": true,
 
-    "StatusList2021Credential": {
+    "StatusListCredential2021": {
       "@id":
         "https://w3id.org/vc/status-list#StatusList2021Credential",
       "@context": {
@@ -16,54 +16,39 @@
       }
     },
 
-    "RevocationList2021": {
+    "StatusList2021": {
       "@id":
-        "https://w3id.org/vc/status-list#RevocationList2021",
+        "https://w3id.org/vc/status-list#StatusList2021",
       "@context": {
         "@protected": true,
 
         "id": "@id",
         "type": "@type",
 
+        "statusPurpose":
+          "https://w3id.org/vc/status-list#statusPurpose",
         "encodedList": "https://w3id.org/vc/status-list#encodedList"
       }
     },
 
-    "RevocationList2021Status": {
+    "StatusListEntry2021": {
       "@id":
-        "https://w3id.org/vc/status-list#RevocationList2021Status",
+        "https://w3id.org/vc/status-list#StatusListEntry2021",
       "@context": {
         "@protected": true,
 
         "id": "@id",
         "type": "@type",
 
+        "statusPurpose":
+          "https://w3id.org/vc/status-list#statusPurpose",
+        "statusListIndex":
+          "https://w3id.org/vc/status-list#statusListIndex",
         "statusListCredential": {
           "@id":
             "https://w3id.org/vc/status-list#statusListCredential",
           "@type": "@id"
-        },
-        "statusListIndex":
-          "https://w3id.org/vc/status-list#statusListIndex"
-      }
-    },
-
-    "SuspensionList2021Status": {
-      "@id":
-        "https://w3id.org/vc/status-list#SuspensionList2021Status",
-      "@context": {
-        "@protected": true,
-
-        "id": "@id",
-        "type": "@type",
-
-        "statusListCredential": {
-          "@id":
-            "https://w3id.org/vc/status-list#statusListCredential",
-          "@type": "@id"
-        },
-        "statusListIndex":
-          "https://w3id.org/vc/status-list#statusListIndex"
+        }
       }
     }
   }
--- a/index.html
+++ b/index.html
@@ -276,7 +276,7 @@ The following sections outlines the data model for this document.
       </p>
 
       <section>
-        <h3>RevocationList2021Status</h3>
+        <h3>StatusListEntry2021</h3>
 
         <p>
 When an <a>issuer</a> desires to enable status information for a
@@ -305,7 +305,39 @@ status list.
             <tr>
               <td>type</td>
               <td>
-The <code>type</code> property MUST be <code>RevocationList2021Status</code>.
+The <code>type</code> property MUST be <code>StatusListEntry2021</code>.
+              </td>
+            </tr>
+            <tr>
+              <td>statusPurpose</td>
+              <td>
+The purpose of the status entry MUST be a string. While the value of the
+string is arbitrary, the following values MUST be used for their intended
+purpose:
+                <table class="simple">
+                  <thead>
+                    <tr>
+                      <th style="white-space: nowrap">Value</th>
+                      <th>Description</th>
+                    </tr>
+                  </thead>
+                  <tbody>
+                    <tr>
+                      <td><code>revocation</code></td>
+                      <td>
+Used to cancel the validity of a <a>verifiable credential</a>. This status is
+not reversible.
+                      </td>
+                    </tr>
+                    <tr>
+                      <td><code>suspension</code></td>
+                      <td>
+Used to temporarily prevent the acceptance of a <a>verifiable credential</a>.
+This status is reversible.
+                      </td>
+                    </tr>
+                  </tbody>
+                </table>
               </td>
             </tr>
             <tr>
@@ -313,7 +345,7 @@ The <code>type</code> property MUST be <code>RevocationList2021Status</code>.
               <td>
 The <code>statusListIndex</code> property MUST be an arbitrary size integer
 greater than or equal to 0, expressed as a string. The value identifies the bit
-position of the revocation status of the <a>verifiable credential</a>.
+position of the status of the <a>verifiable credential</a>.
               </td>
             </tr>
             <tr>
@@ -322,7 +354,7 @@ position of the revocation status of the <a>verifiable credential</a>.
 The <code>statusListCredential</code> property MUST be a URL to a
 <a>verifiable credential</a>. When the URL is dereferenced, the resulting
 <a>verifiable credential</a> MUST have <code>type</code> property that
-includes the <code>StatusList2021Credential</code> value.
+includes the <code>StatusListCredential2021</code> value.
               </td>
             </tr>
           </tbody>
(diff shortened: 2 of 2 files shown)


contexts/v1.jsonld
```diff
@@ -2,7 +2,7 @@
   "@context": {
     "@protected": true,
 
-    "StatusList2021Credential": {
+    "StatusListCredential2021": {
```
dlongley, line 5: This shouldn't be changed -- this is a type of `Credential` not a `Credential2021`, so it should end in just `Credential`.

```suggestion
    "StatusList2021Credential": {
```

contexts/v1.jsonld
```diff
@@ -16,54 +16,39 @@
...
         "type": "@type",
 
+        "statusPurpose":
+          "https://w3id.org/vc/status-list#statusPurpose",
         "encodedList": "https://w3id.org/vc/status-list#encodedList"
       }
     },
 
-    "RevocationList2021Status": {
+    "StatusListEntry2021": {
```
dlongley, line 34: It seems like this should be `StatusList2021Entry`, i.e., everything should be `StatusList2021X`.

contexts/v1.jsonld
```diff
@@ -16,54 +16,39 @@
...
-        "https://w3id.org/vc/status-list#RevocationList2021",
+        "https://w3id.org/vc/status-list#StatusList2021",
       "@context": {
         "@protected": true,
 
         "id": "@id",
         "type": "@type",
 
+        "statusPurpose":
+          "https://w3id.org/vc/status-list#statusPurpose",
```
dlongley, line 29: I'm not sure if this is better than `statusType` ... but maybe. Still considering.

index.html
```diff
@@ -276,7 +283,7 @@ <h2>Data Model</h2>
       </p>
 
       <section>
-        <h3>RevocationList2021Status</h3>
+        <h3>StatusListEntry2021</h3>
```
dlongley, line 286: Here and throughout:

```suggestion
        <h3>StatusList2021Entry</h3>
```

index.html
```diff
@@ -400,25 +361,25 @@ <h3>SuspensionList2021Status</h3>
 The <code>statusListCredential</code> property MUST be a URL to a
 <a>verifiable credential</a>. When the URL is dereferenced, the resulting
 <a>verifiable credential</a> MUST have <code>type</code> property that
-includes the <code>StatusList2021Credential</code> value.
+includes the <code>StatusListCredential2021</code> value.
```
dlongley, line 364: Here and throughout:

```suggestion
includes the <code>StatusList2021Credential</code> value.
```

dlongley (changes requested):

contexts/v1.jsonld
msporny: Yes, considered that. It's strange... can't think of anywhere else that we do this. It creates this new rule that people have to remember and it's not clear when to use one vs. the other. I'd rather we stick with one rule (dates go at the end) rather than (dates sometimes go in the middle -- here's yet another complex VC thing you have to remember to determine when to do what).

contexts/v1.jsonld
msporny: I'm concerned about adding too many `xType` fields... people are going to get confused between that and the regular `type` field.

contexts/v1.jsonld
msporny: Same comment as https://github.com/w3c-ccg/vc-status-list-2021/pull/24/files#r825341914.

index.html
msporny: Same comment as https://github.com/w3c-ccg/vc-status-list-2021/pull/24/files#r825341914.

index.html
msporny: Same comment as https://github.com/w3c-ccg/vc-status-list-2021/pull/24/files#r825341914.

contexts/v1.jsonld
dlongley: The other rule we have always had is that `Credential` is last. This would break that rule.

dlongley: Hmm, I don't think that there's a "new rule" here with date stamps. I think these have already been the rules:

1. If you want to date stamp a thing, `X`, with a year, `YEAR`, do `XYEAR`.
2. If you want a new type of `Credential` to hold / talk about `X` it should be `XCredential`. This means `XYEARCredential` for date stamped `X`'s.

I don't think we're date-stamping the credential, we're date-stamping the status list. Then, as part of how this system works, we also need a credential to put the status list in. That's what the credential is ... a credential that holds a `StatusList2021`. It's not a credential ... from 2021 (or something?) that is holding an unstamped status list. I'm not sure what the date stamp means if it's put at the end here.

I think it's better that the rules be that the date follows the thing it applies to -- which is the status list definition. The fact that there's a credential to go along with a `StatusList2021` is orthogonal, IMO.

contexts/v1.jsonld
dlongley: Yeah, I think this is probably a good approach.

contexts/v1.jsonld
msporny: One thing I considered was just dropping the 2021 and just making it "StatusListCredential". Thoughts?

msporny: > 2\. If you want a new type of `Credential` to hold / talk about `X` it should be `XCredential`. This means `XYEARCredential` for date stamped `X`'s.

In practice, this hasn't been a thing, has it? I'm thinking we could simplify things greatly by eliminating this "rule" (debatable). I don't think it's buying us much... if anything. What is it buying us?

dlongley: @msporny,

> One thing I considered was just dropping the 2021 and just making it "StatusListCredential". Thoughts?

Well, that would mean that every `StatusListCredential` would be expected to use the other `StatusList2021` terms in it. So, -1.

> In practice, this hasn't been a thing, has it? I'm thinking we could simplify things greatly by eliminating this "rule" (debatable). I don't think it's buying us much... if anything. What is it buying us?

If you're referring to `XCredential`, that's been common practice for a long while. If you're referring to `XYEARCredential` (where `X` was date stamped), then it has also been the practice, however, there is only one instance of that practice that I know of: [RevocationList2020Credential](https://w3c-ccg.github.io/vc-status-rl-2020/#revocationlist2020credential).

What it buys us is consistency in naming convention (a credential for `X` is `XCredential` -- whether or not `X` is date stamped) and not squatting on a name like `StatusListCredential` for every possible date-stamped revision of status list in the future.

sbutterfield: Per `CredentialStatusList2017`  and many other types seem to be ubiquitous to me. But this is a credential right? Hm.

My personal feeling is that the whole `YEAR` being part of a naming convention for a Credential is: (a) a little weird, but I get why you might want it ... (b) if present, should serve as some kind of demarcation point for a thing for the purposes of being able to have different versions of it that are "applicable from the time of `YYYY`" ... 
Like, if 'versioning' of a credential type is a worry, wouldn't you just version the schema linked with the credential type instead? 🤷🏻 

Other credentials I've observed all end in `Credential` keyword and I don't think I've seen one with a `YYYY` pattern anywhere in it. I am admittedly new to exploring every credential type though.

Seems like each `XX Data Model` specification would put in place the syntactic requirement so that a naming convention could be somewhat reliable.

contexts/v1.jsonld
```diff
@@ -2,7 +2,7 @@
   "@context": {
     "@protected": true,
```
OR13, line 3: I wonder if just include an `@vocab` would be a better move here.

contexts/v1.jsonld
msporny: > Like, if 'versioning' of a credential type is a worry, wouldn't you just version the schema linked with the credential type instead? 🤷🏻

The challenge is that, at the end of the day, if we're dealing w/ Linked Data... that value with `YYYY` on it expands to a URL. For example, `https://w3id.org/vc/status-list#StatusList2021`... if we get rid of the `YYYY`, then it'll expand to `https://w3id.org/vc/status-list#StatusList`... forever. That means, we can never change what that means semantically, which has consequences... like, what fields are expected to go with an object of that type? That's why we year-stamp some of these things that we're fairly certain are going to change in meaning over the coming decades. A StatusList in 2021 uses a bitstring... but a StatusList in 2030 might use a quantum superposition value. So, StatusList2021 means one thing, and StatusList2030 can mean something else entirely.

Some things, like `VerifiableCredential` are so foundational (and generalized) that we have a high degree of certainty that they're not going to change. However, Ed25519Signature2018 -- we're certain is going to change as 1) we get more and more implementation experience, and 2) as the cryptographic primitives are defeated, over time.

sbutterfield: Interesting. I get you. Seems like `StatusList2021Credential` agrees the "most" with naming conventions I suppose. 
Other options like putting a `/vYYYY/` in the path would cause issues with resolution down the line. 
Seems weird, but I agree that this particular credential type is far from settled.

msporny: > Seems like `StatusList2021Credential` agrees the "most" with naming conventions I suppose.

I count that as 2 for, 1 undecided... would be good to get some more opinions on this... /cc @OR13 @mprorock @mkhraisha @tplooker @mavarley @TallTed @peacekeeper

contexts/v1.jsonld
msporny: Would be good to get some more opinions on this as well. /cc @OR13 @mprorock @mkhraisha @tplooker @mavarley @TallTed @peacekeeper

contexts/v1.jsonld
msporny: Why do you think that, @OR13? If we do that, then everything under StatusList2021 becomes a part of the status list vocabulary. Seems like a dangerous thing to just blanket map any term to the status list vocabulary (this is why `@vocab`, in general, is a bad idea and shouldn't be used).

contexts/v1.jsonld
TallTed: `StatusList2021Credential` being the Credential that holds the 2021 StatusList which is made up of some number of `StatusList2021Entry` Entries *seems* to make the most sense. I think it will need some clear description of how it was arrived at, something like my first sentence, to carry forward clearly into the 2022 or 2025 or 2057 editions...

aljones15: Just so there is more consensus, `XYearCredential` seems fine to me. So `StatusList2021Credential` is ok. The naming convention reminds me of ECMA script standards such as `ES2016` which are working fine in the javascript community.

contexts/v1.jsonld
aljones15: Also agreeing here: `statusPurpose` is a better name than `statusType` which creates ambiguity with numerous of uses of the term `type.`

contexts/v1.jsonld
msporny: Fixed in 01a681de2d2c2b91fc13d5ed6b272ea27254abe4.

index.html
msporny: Fixed in 01a681de2d2c2b91fc13d5ed6b272ea27254abe4.

index.html
msporny: Fixed in 01a681de2d2c2b91fc13d5ed6b272ea27254abe4.

Issue #36, opened 2023-01-08
Specify bit order for status list 


msporny:
We need to add text to the specification that notes that the bit order for bitstring, namely in an array of bits from 0..N, the left most bit is index 0, and the right most bit is index N. This is not to be confused with big-endian vs. little-endian byte order.

dlongley:
Specifically, the left most bit of a bitstring is the first bit (index 0) and the right most bit is the last bit.

Sakurann:
+1. 

> the left most bit of a bitstring is the first bit (index 0) and the right most bit is the last bit (index length - 1). It may be better to describe it this way to avoid confusing byte order with bit order and so on. 

this explanation would be very very helpful. We have faced a case where the issuer's revocation service and the verifiers service had different implementations and were checking the wrong bit for revocation.

TallTed:
> some example images would also be helpful.

Yes, please. I think I understand intent, but I'm not sure.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-24-vcwg-special#section1-4) on 2023-01-24

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.4. Specify bit order for status list  (issue vc-status-list-2021#36)

_See github issue [vc-status-list-2021#36](https://github.com/w3c/vc-status-list-2021/issues/36)._




**Kristina Yasuda:** ten minutes left. one more issue #36.  

**Manu Sporny:** yes, we should do this.  
… we already have examples of implementers interpreting this differently. Ready for PR.  

**Kristina Yasuda:** any objections?.  
… [none were heard].  

> *Brent Zundel:* I would call them processing rules. Once processed, the Verifier can still determine whether the information applies to its internal validation rules..

</details>

msporny:
PR #45 has been raised to address this issue. This issue will be closed once PR #45 has been merged.

msporny:
PR #45 has been merged, closing.

Issue #8, opened 2021-09-28
Bitstring endianness


clehner:
The endianness of the bitstring should be specified.

i.e. Does index "0" of the bitstring refer to the leftmost bit in the first byte (bit mask `0x80`) or to the rightmost bit of the first byte (bit mask `0x01`)

dlongley:
IIRC, current implementations use little endian (index of `0` refers to the leftmost bit) and +1 to making this clear in the spec and including diagrams.

dlongley:
Hmm, no, just checked and looks like current implementations use big endian (index of `0` refers to the right most bit in the first byte (`0x01`). We will want to get this cleared up and make a choice.

dlongley:
We've updated our implementations to ensure that both the bit order and the byte order are consistently little endian. We should specify this in the spec and fix any examples that might suggest otherwise.

dlongley:
See issue #36.

OR13:
suggest closing as a dupe after #36

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-24-vcwg-special#section1-7) on 2023-01-24

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.7.  (issue vc-status-list-2021#8)

_See github issue [vc-status-list-2021#8](https://github.com/w3c/vc-status-list-2021/issues/8)._




**Kristina Yasuda:** Issue 8.  
… Looks like debates about issue..  

> *Dave Longley:* this is a duplicate.

**Manu Sporny:** This looks like a dupe.  

**Kristina Yasuda:** Ok, we'll close it as a dup.  
… Thanks everyone. That was a productive discussion.  
… Looking forward to reviewing PRs in main call.  
… See you all tomorrow at the same time.  

---

</details>

PR #45, opened 2023-01-29
Add section about bitstring encoding.

msporny:
This PR adds a normative statement regarding the encoding format for the status list bitstring as well as a Security Considerations section related to not properly encoding or decoding the bitstring. It addresses issues #8 and #36.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-status-list-2021/pull/45.html" title="Last updated on Feb 4, 2023, 9:54 PM UTC (5e6cfa8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-status-list-2021/45/ca50a56...5e6cfa8.html" title="Last updated on Feb 4, 2023, 9:54 PM UTC (5e6cfa8)">Diff</a>


index.html
```diff
@@ -472,7 +472,13 @@ <h3>StatusList2021Credential</h3>
...
 the GZIP-compressed [[RFC1952]], base-64 encoded [[RFC4648]] bitstring values
 for the associated range of <a>verifiable credential</a> status values. The
-uncompressed bitstring MUST be at least 16KB in size.
+uncompressed bitstring MUST be at least 16KB in size. The bitstring MUST be
+encoded as a big-endian value with a start index value of zero
+(<code>0</code>) located at the left-most bit in the bitstring and the
+end index of one less than the length of the bitstring
+(<code>bitstring_length - 1</code>) and located at the right-most bit in the
+bitstring. Further information on bitstring encoding can be found in Section
+<a href="#bitstring-encoding"></a>.
```
dlongley, line 481:
```suggestion
uncompressed bitstring MUST be at least 16KB in size. The bitstring MUST be
encoded such that the first index, with a value of zero
(<code>0</code>), is located at the left-most bit in the bitstring and the
last index, with a value of one less than the length of the bitstring
(<code>bitstring_length - 1</code>), is located at the right-most bit in the
bitstring. Further information on bitstring encoding can be found in Section
<a href="#bitstring-encoding"></a>.
```

index.html
```diff
@@ -724,9 +730,35 @@ <h2>Security Considerations</h2>
...
+
+      <p>
+It is critical that implementers pay particular attention to the way that they
+encode and decode bitstrings. Failure to do so can result in a failure to check
+the correct position in a bitstring. Checking an incorrect bitstring index can
+lead to failures such as mistaking a revoked credential for one that is not
+revoked. As stated in Section <a href="#statuslist2021credential"></a>,
+bitstrings are encoded as big-endian values with the index starting at the
+left-most index of the bitstring array. The diagram below demonstrates the
+proper layout for an uncompressed bitstring.
```
dlongley, line 744:
```suggestion
It is critical that implementers pay particular attention to the way that they
encode and decode bitstrings. Failure to do so can result in checking the
wrong bitstring index for a given credential, leading to a misinterpretation
of its present state (e.g., mistakening a revoked status for an unrevoked
status). As stated in Section <a href="#statuslist2021credential"></a>,
bitstrings are encoded such that the first (zeroth) index refers to the
left-most bit of the bitstring array. The diagram below demonstrates the
proper layout for an uncompressed bitstring.
```

index.html
```diff
@@ -724,9 +730,35 @@ <h2>Security Considerations</h2>
...
+      </p>
+
+      <figure id="bitstring">
+        <img style="margin: auto; display: block; width: 75%;"
+             src="diagrams/bitstring-layout.svg" alt="a diagram showing two
+             wide rectangles shown side-by-side. Each rectangle is partitioned
+             into eight boxes to represent the 8 bits in a byte. The left
+             rectangle is labeled 'First byte' and the right rectable is
```
dlongley, line 752:
```suggestion
             rectangle is labeled 'First byte' and the right rectangle is
```

dlongley (approved):
Approving w/suggestions. Thanks!

index.html
TallTed: `s/mistakening/mistaking/` and I'm good with this change.

index.html
TallTed: We understand `bitstring_length - 1` as written.  Later readers will benefit by a contrived example that shows, for instance, `bitstring_length = 17000` and `last a/k/a right-most index = 16999`

TallTed (changes requested):
I'll be good after the spelling correction, and addition of an example, as noted on @dlongley's suggestions, with which I otherwise concur.

mprorock (changes requested):
minor improvements as noted by @TallTed and @dlongley - otherwise great

OR13 (approved):

OR13:
(assuming comments by others are addressed).

dwight-holman:
Can we use "most significant" and "least significant" instead of "left" and "right"?  Within a byte, these aren't affected by endianess and would be completely unambiguous. 

I'd also add a note that this goes contrary to many implementations of bitstrings, so care needs to be taking when using standard libraries, but that might be outside the scope of this document.

dwight-holman:
Also note that many existing implementations of revocation lists have this reversed.

OR13:
> It'd be nice if there was a way for existing implementations to distinguish between the old ordering and the new in existing VCs.

StatusList2023 here we come :)

msporny:
> Can we use "most significant" and "least significant" instead of "left" and "right"? Within a byte, these aren't affected by endianess and would be completely unambiguous.

"Completely unambiguous" is a very difficult bar to reach when you're involving developers from a variety of different backgrounds... MSB/LSB and then throw endianness into the mix... it gets confusing quickly (as demonstrated by the thread below):

https://github.com/scott-griffiths/bitstring/issues/156

That said, yes, let's try to find some language that makes sense to everyone while understanding that it becomes confusing quickly when you mix in MSB/LSB and endianness.

> Also note that many existing implementations of revocation lists have this reversed. 

What are the existing implementations that you know about that have this reversed? As far as we know, there are only a handful of experimental implementations for status list 2021, nothing has been shipped to production yet.

> Edit: It'd be nice if there was a way for existing implementations to distinguish between the old ordering and the new in existing VCs.

Or, the existing implementations were implementing a spec that was not stable and shouldn't expect that things won't change. From the "Status of the Document" section:

> This document is experimental and is undergoing heavy development. It is inadvisable to implement the specification in its current form. An [experimental implementation](https://github.com/digitalbazaar/vc-status-list) is available.

Now, if status list 2021 has shipped in some sort of production capacity, we do need to take that into account. Do you know of any status list 2021 implementations that have shipped to production (against the guidance in the spec)?

index.html
msporny: Fixed.

index.html
```diff
@@ -724,9 +730,35 @@ <h2>Security Considerations</h2>
...
+             rectangle is labeled 'First byte' and the right rectangle is
+             labeled 'Last byte'. The left rectangle's left-most bit box is
+             pointed to by an arrow labeled 'index: 0'. The right rectangle's
+             right-most bit box is pointed to by an arrow labeled
+             'index: length - 1'.">
+        <figcaption style="text-align: center;">
+          A visual depiction of the bitstring layout.
+        </figcaption>
+      </figure>
+    </section>
```
msporny, line 761:
```suggestion
      </figure>
      
      <p>
For example, if a bitstring is 200,000 bits in size, the first index will be 0, and the last index will be 199,999.
      </p>
    </section>
```

msporny: @TallTed does this address your concern? I used a number greater than 17,000 because implementers are not supposed to use bitstrings shorter than 131,072 bits (16KB is the minimum allowable herd size).

dlongley: Maybe just use `131,072 bits in size` ... `last index will be `131,071` to keep with the 16KB minimum. Using some size that isn't a power of 2 may lead to more questions. In fact, we might want to explicitly say that all the bit string lengths must be a power of 2 to keep implementations simple -- and that that doesn't mean all bits must be assigned to something, just that the total length will always a power of 2.

msporny: Updated to use 131,072 bits... feels a bit less understandable than 200,000 vs. 199,999, but meh... not that much less?

TallTed: Works for me.

msporny:
> Given the mismatch between common bitstring implementations and the one being specified, it's also going to be more likely to cause the security concerns described.

Ok, @dwight-holman, I'm taking your comment as blocking this PR from being merged until there can be further discussion.

The Microsoft implementation seems to use `RevocationList2021Status`, which is a different spec, so doesn't seem to apply to this spec (that spec can use a different bitstring encoding because it uses a different type). It's not clear to me how they implemented the bitstring. We can ask them on the upcoming call. The same goes for Ping.

> Re: implementations of bitstrings, IIRC Go, Python, Java and Typescript use the other order.

Do you have links so that the rest of us don't have to go off to hunt those things down?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-31-vcwg-special#section1-1) on 2023-01-31

- no resolutions were taken

#### 1.1. Add section about bitstring encoding. (pr vc-status-list-2021#45)

_See github pull request [vc-status-list-2021#45](https://github.com/w3c/vc-status-list-2021/pull/45)._




**Manu Sporny:** there 2 PRs follow up from discussions last week.  
… they cover bitstring and dereferencing..  
… it adds an image to talk about encoding, and some language..  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-02-01-vcwg#section3-6) on 2023-02-01

- no resolutions were taken

#### 3.6. Add section about bitstring encoding. (pr vc-status-list-2021#45)

_See github pull request [vc-status-list-2021#45](https://github.com/w3c/vc-status-list-2021/pull/45)._




**Manu Sporny:** PR 45 adds a section about bitstring encoding, how exactly is the bitstring encoded. There's a PR that clarifies this and adds examples. I don't think there is a lot of disagreement..  

</details>

dwight-holman:
(I'm going to use Lsb0 and Msb0 as shorthand for if 0x01 has index 0 or index 7 set.)

Go: https://play.golang.com/p/P0fSWgMsnIa
Standard library only has ASN.1 bit strings (Msb0), so I checked the library being used by TDB's [StatusList2021 implementation](https://github.com/TBD54566975/ssi-sdk/blob/6519a59e9e5e81b5a029324c316c65a03a381fc3/credential/status/statuslist2021.go) and it's Lsb0.

Java has [BitSet](https://docs.oracle.com/javase/8/docs/api/java/util/BitSet.html#valueOf-byte:A-) which is Lsb0.

Python has nothing in the standard library, but it does look like the [most popular library](https://pypi.org/project/bitarray/) does Msb0 by default.    

Rust's [BitArray](https://docs.rs/bitvec/latest/bitvec/array/struct.BitArray.html) defaults to Lsb0 but does support both.

Javascript/typescript, no standard library implementation.  Popular [bitset](https://www.npmjs.com/package/bitset) library is Msb0.

I'm having a hard time finding the ancient OSS revocation implementation I cross-checked ours with in the beginning, but I do remember it was RevocationList (not StatusList) so it's old enough to discount.

There are a few JS/TS implementations of StatusList2021 on Github, though, so I checked those:

- [Energywebfoundation](https://github.com/energywebfoundation/ew-credentials/blob/c54672118a2687146e8421d3779208a04ee191c0/packages/vc-verification/src/revocation-verification.ts) doesn't seem to check the actual revocation list yet.
- [CentreHQ Verite](https://github.com/centrehq/verite/blob/5ab317480d5b82e5872ff8bdfdc88499ff68ac17/packages/verite/lib/issuer/revocation.ts) uses [bit-buffers](https://github.com/m2-labs/bit-buffers/), which appear to be Msb0 after decoding some of the test fixtures.
- [Trustify](https://github.com/TrustifyX/trustify/blob/baa2fd2dbfd57d13092165e6c03e893b1f7b37b6/packages/frontend/lib/issuer/revocation.ts) uses the same bit-buffers library, Msb0.
- [GravityID](https://github.com/GravityID/status-manager/blob/master/src/bitstring.ts) rolled their own Lsb0.
- [CheqD](https://github.com/cheqd) was also a match for StatusList2021 while searching, but doesn't seem to have a public implementation yet.

Not as clear of a bias towards Lsb0 as I thought!  In the absence of other factors, I'd guess an implementer would default to Msb0 in Python and JS; with Lsb0 being the default for Java and Rust.

dwight-holman:
Oh!  Before I forget, all of the bitstrings from the JS implementations use `pako` under the hood, which is a zlib implementation.  The outputs from that library lack the member headers mandated by the GZip standard being referenced (RFC1952), so none of them are in compliance anyways. 😄

OR13:
@dwight-holman yes, thank your for mentioning `pako`... I have hit this exact issue several times... it can also impact content addressing... compression must be handled with extra care, I suggest filing a separate issue on it.

dlongley:
@dwight-holman,

Thank you for the bitstring indexing order research!

As for the comment about `pako` -- are you saying that if people use its `gzip()` API that the output is not compliant with RFC 1952? Or are you saying if they call the `deflate()` API erroneously then it's not compliant? Do you have an example demonstrating the problem?

msporny:
> Thank you for the bitstring indexing order research!

+1, THANK YOU. That was excellent and is exactly the sort of thing we need to make progress. It's too bad that there wasn't a clear indication one way or the other. :P 

It does fit the "there is no 'obvious' way to do this sort of encoding" suggestion... I guess where we are right now is if anyone would object if we merged this PR as-is? Whatever we do, it looks like half of the implementations would get it wrong (so, we clearly need guidance here).

dwight-holman:
> Or are you saying if they call the deflate() API erroneously then it's not compliant? Do you have an example demonstrating the problem?

Per the RFC, the first three bytes of a valid message are 0x1f 0x8b 0x08, and the `bit-buffers` values lack those.  Since the first three bytes encode to `H4sI` it's a quick way to check validity.

BitBuffer's toBitstring() uses pako's deflate, giving strings like "eJwTYEAFAAEQABE=" lacking the gzip headers.

pako's gzip() looks fine, based on poking the GravityId tests.

So the issue isn't pako: it's bit-buffers and all those different StatusList2021 implementations can get fixed in one place.

OR13:
Maybe include a citation for the encoding to any relevant RFCs?

- https://datatracker.ietf.org/doc/html/rfc3642

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-02-08-vcwg#section3-3) on 2023-02-08

- no resolutions were taken

#### 3.3. Add section about bitstring encoding. (pr vc-status-list-2021#45)

_See github pull request [vc-status-list-2021#45](https://github.com/w3c/vc-status-list-2021/pull/45)._




**Manu Sporny:** vc status list has 3 PRs.  
… good discussion about bitstring encoding.  
… no one way to encode bitstrings.  
… need to specify how to do this in space.  
… going to ask if dwight is objecting.  
… implementers need to know to check implementations.  

</details>

msporny:
Normative, multiple reviews, changes requested and made, issues raised for remaining concerns, no objections, merging.

PR #60, opened 2023-04-21
Add JSON Web Token Examples

OR13:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-status-list-2021/pull/60.html" title="Last updated on Apr 21, 2023, 10:39 PM UTC (bea57c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-status-list-2021/60/127d7f0...bea57c4.html" title="Last updated on Apr 21, 2023, 10:39 PM UTC (bea57c4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -859,5 +859,94 @@ Write i18n considerations.
 
   </section>
 
+  <section>
+    <h2>Appendix</h2>
+    <section>
+      <h2>JSON Web Token Examples</h2>
+      <pre class="example nohighlight" title="JSON Web Token Status List">
+eyJhbGciOiJFUzI1NiIsImlzcyI6ImRpZDpleGFtcGxlOjEyMyIsImtpZCI6IiMwIiwidHlwIjoidmMrbGQrand0IiwiY3R5IjoidmMrbGQranNvbiJ9.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiLCJodHRwczovL3czaWQub3JnL3ZjL3N0YXR1cy1saXN0LzIwMjEvdjEiXSwiaWQiOiJodHRwczovL3ZlbmRvci5leGFtcGxlL2NyZWRlbnRpYWxzL3N0YXR1cy8zIiwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIlN0YXR1c0xpc3QyMDIxQ3JlZGVudGlhbCJdLCJpc3N1ZXIiOiJkaWQ6ZXhhbXBsZToxMjMiLCJ2YWxpZEZyb20iOiIyMDIxLTA0LTA1VDE0OjI3OjQwLjAwMFoiLCJjcmVkZW50aWFsU3ViamVjdCI6eyJpZCI6Imh0dHBzOi8vdmVuZG9yLmV4YW1wbGUvY3JlZGVudGlhbHMvc3RhdHVzLzMjbGlzdCIsInR5cGUiOiJTdGF0dXNMaXN0MjAyMSIsInN0YXR1c1B1cnBvc2UiOiJzdXNwZW5zaW9uIiwiZW5jb2RlZExpc3QiOiJINHNJQUFBQUFBQUFBMk1BQUkzdkF0SUJBQUFBIn19.eRnHulruzCNFAD4n3PlAkn7hPouynDvMZxmID8bP0hXCZN3_CswKyIOFSyndpD6EhJfET7j5vVEuVrZXnnC_2A
+      </pre>
+      <pre class="example nohighlight" title="JSON Web Token Status List (Verified)">
+{
+  "protectedHeader": {
+    "alg": "ES256",
+    "iss": "did:example:123",
+    "kid": "#0",
+    "typ": "vc+ld+jwt",
+    "cty": "vc+ld+json"
+  },
+  "payload": {
+    "@context": [
+      "https://www.w3.org/ns/credentials/v2",
+      "https://w3id.org/vc/status-list/2021/v1"
+    ],
+    "id": "https://vendor.example/credentials/status/3",
+    "type": [
+      "VerifiableCredential",
+      "StatusList2021Credential"
+    ],
+    "issuer": "did:example:123",
+    "validFrom": "2021-04-05T14:27:40.000Z",
+    "credentialSubject": {
+      "id": "https://vendor.example/credentials/status/3#list",
+      "type": "StatusList2021",
+      "statusPurpose": "suspension",
+      "encodedList": "H4sIAAAAAAAAA2MAAI3vAtIBAAAA"
+    }
+  }
+}
+      </pre>
+      <pre class="example nohighlight" title="JSON Web Token Verifiable Credential With Status">
+eyJhbGciOiJFUzI1NiIsImlzcyI6ImRpZDpleGFtcGxlOjEyMyIsImtpZCI6IiMwIiwidHlwIjoidmMrbGQrand0IiwiY3R5IjoidmMrbGQranNvbiJ9.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiLCJodHRwczovL3czaWQub3JnL3ZjL3N0YXR1cy1saXN0LzIwMjEvdjEiXSwiaWQiOiJodHRwOi8vZXhhbXBsZS5jb20vY3JlZGVudGlhbHMvMTg3MiIsInR5cGUiOlsiVmVyaWZpYWJsZUNyZWRlbnRpYWwiLCJOZXdDcmVkZW50aWFsVHlwZSJdLCJpc3N1ZXIiOnsiaWQiOiJkaWQ6ZXhhbXBsZToxMjMiLCJ0eXBlIjpbIk9yZ2FuaXphdGlvbiIsIk9yZ2FuaXphdGlvblR5cGUiXX0sInZhbGlkRnJvbSI6IjIwMTAtMDEtMDFUMTk6MjM6MjRaIiwiY3JlZGVudGlhbFN0YXR1cyI6eyJpZCI6Imh0dHBzOi8vZXhhbXBsZS5jb20vY3JlZGVudGlhbHMvc3RhdHVzLzMjNCIsInR5cGUiOiJTdGF0dXNMaXN0MjAyMUVudHJ5Iiwic3RhdHVzUHVycG9zZSI6InN1c3BlbnNpb24iLCJzdGF0dXNMaXN0SW5kZXgiOiI0Iiwic3RhdHVzTGlzdENyZWRlbnRpYWwiOiJodHRwczovL2V4YW1wbGUuY29tL2NyZWRlbnRpYWxzL3N0YXR1cy8zIn0sImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImlkIjoiZGlkOmV4YW1wbGU6NDU2IiwidHlwZSI6WyJQZXJzb24iLCJKb2JUeXBlIl0sImNsYWltTmFtZSI6ImNsYWltVmFsdWUifX0.krP-9305AZEUu686iw_DfakBihBYPbyXDqQ51n2eYLJolg04k0RFS7g-JCKbpRkUclqQKrN3OBs2SQdcMgGpWg
+      </pre>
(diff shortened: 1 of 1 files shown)


OR13:
I generated these examples from this implementation: https://github.com/transmute-industries/vc-jwt-status-list

I plan to add similar examples to this test suite:

https://github.com/transmute-industries/vc-jwt-test-suite

mprorock (approved):
Thank you @OR13 !

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
+eyJhbGciOiJFUzI1NiIsImlzcyI6ImRpZDpleGFtcGxlOjEyMyIsImtpZCI6IiMwIiwidHlwIjoidmMrbGQrand0IiwiY3R5IjoidmMrbGQranNvbiJ9.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiLCJodHRwczovL3czaWQub3JnL3ZjL3N0YXR1cy1saXN0LzIwMjEvdjEiXSwiaWQiOiJodHRwOi8vZXhhbXBsZS5jb20vY3JlZGVudGlhbHMvMTg3MiIsInR5cGUiOlsiVmVyaWZpYWJsZUNyZWRlbnRpYWwiLCJOZXdDcmVkZW50aWFsVHlwZSJdLCJpc3N1ZXIiOnsiaWQiOiJkaWQ6ZXhhbXBsZToxMjMiLCJ0eXBlIjpbIk9yZ2FuaXphdGlvbiIsIk9yZ2FuaXphdGlvblR5cGUiXX0sInZhbGlkRnJvbSI6IjIwMTAtMDEtMDFUMTk6MjM6MjRaIiwiY3JlZGVudGlhbFN0YXR1cyI6eyJpZCI6Imh0dHBzOi8vZXhhbXBsZS5jb20vY3JlZGVudGlhbHMvc3RhdHVzLzMjNCIsInR5cGUiOiJTdGF0dXNMaXN0MjAyMUVudHJ5Iiwic3RhdHVzUHVycG9zZSI6InN1c3BlbnNpb24iLCJzdGF0dXNMaXN0SW5kZXgiOiI0Iiwic3RhdHVzTGlzdENyZWRlbnRpYWwiOiJodHRwczovL2V4YW1wbGUuY29tL2NyZWRlbnRpYWxzL3N0YXR1cy8zIn0sImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImlkIjoiZGlkOmV4YW1wbGU6NDU2IiwidHlwZSI6WyJQZXJzb24iLCJKb2JUeXBlIl0sImNsYWltTmFtZSI6ImNsYWltVmFsdWUifX0.krP-9305AZEUu686iw_DfakBihBYPbyXDqQ51n2eYLJolg04k0RFS7g-JCKbpRkUclqQKrN3OBs2SQdcMgGpWg
+      </pre>
+      <pre class="example nohighlight" title="JSON Web Token Verifiable Credential With Status (Verified)">
+{
+  "suspension": false,
```
msporny, line 904: What is this property for?

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
+  },
+  "payload": {
+    "@context": [
+      "https://www.w3.org/ns/credentials/v2",
+      "https://w3id.org/vc/status-list/2021/v1"
+    ],
+    "id": "http://example.com/credentials/1872",
+    "type": [
+      "VerifiableCredential",
+      "NewCredentialType"
```
msporny, line 920: Undefined term, please use a context that defines the term given that this is an example in a specification.

Suggest that we define a `https://www.w3.org/ns/credentials/examples/v2` context that maps everything to an examples vocabulary document, much like undefined terms, but specifically state that this property is an example property and only used for examples. 

Either that, or use a real VC (traceability vocab would be fine).

msporny (changes requested):
Overall, +1 to this PR. There is one question and one request for change before merge.

mkhraisha (approved):
+1 pending manu's comments

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
+    ],
+    "issuer": {
+      "id": "did:example:123",
+      "type": [
+        "Organization",
+        "OrganizationType"
```
dlongley, line 926: These are also undefined, right?

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
+      "statusListIndex": "4",
+      "statusListCredential": "https://example.com/credentials/status/3"
+    },
+    "credentialSubject": {
+      "id": "did:example:456",
+      "type": [
+        "Person",
+        "JobType"
+      ],
+      "claimName": "claimValue"
```
dlongley, line 943: More undefined things.

index.html
OR13: -1 to adding industry specific contexts to examples

-1 to inventing a new examples context (vocab expands these terms automatically).

index.html
OR13: They are not "undefined", they are "vocab defined".

index.html
OR13: More vocab defined things.

index.html
OR13: This is just a json member in the verification result reflecting the status bit for the credential.

It is an implementation artifact, I can remove it.

AFAIK, there is no consensus on how status impacts verification, or how to express that in a "verified JWT".

SafeJWT APIs tend to return an object with protected header and decoded and parsed payload after verification, I added this object member to that to communicate status, but it could just as easily be left to verifier to decide how to service the status check in a verification result.

Last I checked, data integrity does a similar thing for `verified` and status, but it's possible that has changed.

dlongley:
@OR13,

I think it's important that we have "5 star" or "best practice" (or whatever we want to call these) examples that only use non-issuer-specific vocabularies first -- and we call them out as the preferred thing to do. We do not want to encourage everyone to increase the number of potentially "opaque" (or rejected!) VCs in digital wallets.

We can also have examples that show issuer-dependent terms, but we should say that those come with an additional burden, e.g., of digital wallets having to understand every issuer's definition of a term vs. just a single definition from an independent vocab that works across multiple issuers.

So, these things are not the same and just a matter of preference with no external consequences. Issuer-dependent terms can potentially generate market failures. There's significant burden added for "issuer-dependent" terms that should be understood from the three party model perspective. That approach preferences popular issuers by shaping markets because the increased implementation burden on digital wallet providers results in resources being allocated only to these popular issuers. Among of things, this can lead to a repeat of the NASCAR problem and difficulty with smaller issuers being able to get support. We should not suggest, as a first example, that people use "issuer-dependent" terms.

selfissued (approved):
I think it makes sense to have VC-JWT examples.

decentralgabe (approved):
This aligns with the representation [we are using](https://github.com/TBD54566975/ssi-sdk/blob/main/credential/status/statuslist2021.go) in our [service software](https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/credential/service.go#L304) and I'm supportive of this PR.

index.html
mprorock, line 904:
```suggestion
```

index.html
mprorock, line 926:
```suggestion
```

index.html
mprorock, line 920:
```suggestion
```

mprorock:
@dlongley do the suggestions i made above resolve your concerns?

mprorock:
@msporny if the suggestions made above are merged, do they fix your concerns?

dlongley:
> @dlongley do the suggestions i made above resolve your concerns?

No, because issuer-dependent terms are still used as the core example. I think such terms should only be used as auxiliary and appropriately contextualized examples as I mentioned on the call because of their obvious disadvantages for the ecosystem as a whole -- and because developers often use examples (and only examples) as contact points in specifications.

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
+    "alg": "ES256",
+    "iss": "did:example:123",
+    "kid": "#0",
+    "typ": "vc+ld+jwt",
+    "cty": "vc+ld+json"
+  },
+  "payload": {
+    "@context": [
+      "https://www.w3.org/ns/credentials/v2",
+      "https://w3id.org/vc/status-list/2021/v1"
```
dlongley, line 881:
```suggestion
      "https://www.w3.org/ns/credentials/v2",
      "https://www.w3.org/ns/credentials/examples/v2",
      "https://w3id.org/vc/status-list/2021/v1"
```

index.html
```diff
@@ -859,5 +859,94 @@ <h2>Internationalization Considerations</h2>
...
```
dlongley, line 915:
```suggestion
      "https://www.w3.org/ns/credentials/v2",
      "https://www.w3.org/ns/credentials/examples/v2",
      "https://w3id.org/vc/status-list/2021/v1"
```

dlongley (changes requested):
Providing my previous comment as a github suggestion.

msporny:
@dlongley wrote:
> I've provided my previous comment as a github suggestion.

If the change requests suggested by @dlongley are merged, I would be a +1 for merging this example.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-24-vcwg#section1-6) on 2023-05-24

- no resolutions were taken

#### 1.6. Add JSON Web Token Examples (pr vc-status-list-2021#60)

_See github pull request [vc-status-list-2021#60](https://github.com/w3c/vc-status-list-2021/pull/60)._




**Orie Steele:** We seem to have a difference of opinion regarding adding examples to the specification. And I would like to merge the examples as-is. I don't agree with the perspective that every example in our spec needs multiple context values.  
… I don't agree with that and in light of bundling the Data Integrity context.  
… This adds informative examples, can we merge this, please, thank you.  

> *Greg Berstein:* Please!

**Manu Sporny:** -1 to merge until the WG discusses it. With all the examples it keeps coming up and we need to have a consistent way that we're doing examples and figure out what it is so the editors can apply that uniformly across all the specs.  

> *Phillip Long:* is this a special topics call?

**Kristina Yasuda:** I'm up for spending 5-10 minutes on this call on this.  

> *Orie Steele:* lets remove the DataIntegrityProof RDF class from the core context... and make the examples include multiple contexts.

**Manu Sporny:** I don't know if we'll get through it in 5 minutes. The proposal is to just use the example in the core spec.  
… Use it as an example across all the specs that need to use an example. That way we're just using an example that we've agreed to use across all the specs and we don't need to do variations of that in every other specification.  

_See github pull request [vc-data-model#1091](https://github.com/w3c/vc-data-model/pull/1091)._




**Kristina Yasuda:** Are you fundamentally not ok with the example, Manu?  

**Manu Sporny:** Yes.  

**Kristina Yasuda:** Anyone else who has strong opinions about what examples we use?  

**Orie Steele:** All the examples conformant to the core data model.  

**Michael Prorock:** My question is along the same line... what precisely is wrong with the examples?  
… At least by my read and going and running the example code it all checks out fine.  

**Manu Sporny:** One of the things is that we have a suspension false entry in here, I don't know where that property came from, I think Orie said that's an artifact in the PR and it's not in the data model.  
… There's a concern over the use of the undefined / issuer-independent context / vocab.  
… Now we're using the issuer-dependent terms in the example.  
… There are a number of people that said in the other PR that's not a good practice.  
… We've had people say that they would rather the examples use the examples context instead of the issuer-defined one.  
… I don't know why we're doing in status list when core does something else.  

**Dave Longley:** A lot of developers will come along and look at examples to get things started, and there are such things as good examples and bad examples.  

> *Orie Steele:* IMO a good example is one that uses the minimal required fields.  
> it is bad form to include optional stuff in example, and bad form to include DataIntegrityProof in the core v2 context.

**Dave Longley:** You can create examples that are bad examples, which would be conformant, but wouldn't look good. It's not enough to say "We created an example and it validates so it's fine." If we get people started off on he wrong foot, or encouraging market failures, we're making mistakes. We should not be flippant about the examples we put in the spec. They're important, developers will look at these examples and maybe not the text.  
… Saying "this verifies" is not enough, these are entry points into the spec and possibly exit points.  

**Kristina Yasuda:** Sounds like we're leaving this issue open for now.  
… Anything on the BBS cryptosuite stuff?  

</details>

OR13:
Closing the PR, now that https://github.com/w3c/vc-status-list-2021/pull/67 is up, we expect digital bazaar to provide complete examples for VC-JWT using the plugin.

I intend to block CR for this item until working examples for both securing formats are in the spec.

Issue #17, opened 2022-01-27
Add vocabulary document


msporny:
@OR13 wrote:
> No context changes that include dead links.

This specification needs a corresponding vocabulary document to be created.

msporny:
PR #105 has been raised to address this issue. This issue will be closed once PR #105 has been merged.

msporny:
PR #105 has been merged, closing.

PR #100, opened 2023-12-27
Rename "status" to "message" for `statusPurpose` feature.

msporny:
This PR renames the awkwardly named "status" "statusPurpose" field to "message" since the "statusPurpose" is about conveying arbitrary messages to verifiers. This PR will close out issue #42 if merged.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/100.html" title="Last updated on Dec 29, 2023, 11:52 PM UTC (94fb1cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/100/843b1d6...94fb1cc.html" title="Last updated on Dec 29, 2023, 11:52 PM UTC (94fb1cc)">Diff</a>

Issue #81, opened 2023-07-28
Vocabulary definitions


iherman:
(This issue is spawn from https://github.com/w3c/vc-status-list-2021/pull/69#issuecomment-1653831202; it was deemed preferable to have it as a separate issue.)

This document yields a number of terms that must be incorporated into the credentials vocabulary at some point. I hope I am not mistaken with this list:

- **Classes:**
    - StatusListEntry
    - StatusListCredential
    - StatusList
- **Properties:**
    - statusPurpose
      - range: xsd:string
      - domain: StatusListEntry or StatusList
    - statusListIndex
      - range: xsd:string
      - domain: StatusListEntry
    - statusListCredential
      - range: StatusListCredential
      - domain: StatusListEntry
    - encodedList
      - range: xsd:string
      - domain: StatusList

I presume that these terms will be formally defined by this specification (which is perfectly fine from my point of view). However, to anticipate this (following https://github.com/w3c/vc-data-model/pull/1209) there should be clear anchors that provide the referencable URLs for the vocabulary entry into this document. At the moment, there is a URL for`StatusListEntry` and `StatusListCredentials`, but all the others are missing. It may be as simple as providing an `@id` attribute on the table cell elements for the properties, but this will miss the reference for `StatusList` (which is "hidden" in the table row for `credentialSubject.type`).

iherman:
(Copying the answer of @msporny from https://github.com/w3c/vc-status-list-2021/pull/69#issuecomment-1653929062 for an easier discussion.)

@msporny:

> We should probably have a different vocabulary for credential status. It's not required, but it might be the best way to structure things (separation of vocabulary purposes... where the credential vocabulary is for core VCDM things... and extensions get their own vocabularies)? Note that we will need to produce a separate JSON-LD context for status list because of people wanting to pull it into V1 VCs or other software systems (possibly).

iherman:
@msporny 

> We should probably have a different vocabulary for credential status. It's not required, but it might be the best way to structure things (separation of vocabulary purposes... where the credential vocabulary is for core VCDM things... and extensions get their own vocabularies)? 

I am a little bit worried to have too many mini-vocabularies that might become a maintenance nightmare. Should we also separate each cyryptosuite, in the case they add terms, into their own vocabulary? The same for JSON Schemas?

I am not worried if the credential vocab covers different areas. It is the same family of specs anyway...

msporny:
[...]

Issue #73, opened 2023-06-21
Unify design of status lists to support multi-bit values


msporny:
PR https://github.com/w3c/vc-status-list-2021/pull/65 added a backwards-compatible mechanism for binary status lists and multi-value status lists. During the discussion of the PR, there were multiple concerns raised around `ttl`, verification and validation and how the status list might affect those processes, the data formats utilized, the differences between binary and multiple bits per status. It was suggested that perhaps a re-alignment and unification of the design to support the use cases more generally would be useful. This issue tracks that design discussion.

msporny:
Starting us off with a proposed unified design, example first:

```
{
  "id": "https://example.com/status/3#list",
  "type": "StatusList2021",
  "statusField": [{
    "statusBitOffset": 0, 
    "statusPurpose": "revoked",
    "activeStatusMessage": "This credential has been revoked by the issuer.",
    "inactiveStatusMessage": "This credential has been revoked by the issuer."
  }, {
    "statusBitOffset": 1, 
    "statusPurpose": "suspended",
    "activeStatusMessage": "This credential has been suspended by the issuer.",
    "inactiveStatusMessage": "This credential has not been suspended by the issuer."
  }, 
      ...
  ],
  "encodedList": "H4sIAAAAAAAAA-3BMQEAAADAAAAAAAAAAAAAAAAAAAAAAAAAAAIC3AYbSVKsAQAAA"
}
```

* Establishes a new `statusField` property, which contains the information for each field in the bitstring associated with each entry (including messages -- this is somewhat analogous to `statusMessages` in PR #65).
* Moves `statusPurpose` into `statusField`.
* Moves away from hexadecimal notation for status bit position to just a plain integer offset (so we can avoid hex to integer conversion... since we're trying to get to an integer bit offset anyway).
* Adds messages for active and inactive bits.
* Removes the `size` parameter, as it's not needed (you just pick the max statusBitOffset and add one to it's value to get the size.
* Removes `reference`; do we need it anymore if there is a description for every field?
* Removes `ttl`; can we get away with just `validUntil`? 
* Removes the need for `undefined` fields... just leave a gap in your bit offsets if you want to leave fields undefined.

Questions that are left unanswered:

* How is verification and validation affected by the above?  
* We need to write how having multiple bits per status negatively impacts privacy by enabling new correlation vectors?
[...]

PR #105, opened 2023-12-28
Add formal vocabulary definitions.

msporny:
This PR is a first pass at attempting to address issue #17 by formally defining the vocabulary for BitstringStatusList.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/105.html" title="Last updated on Jan 13, 2024, 4:23 PM UTC (b54d651)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/105/479c42d...b54d651.html" title="Last updated on Jan 13, 2024, 4:23 PM UTC (b54d651)">Diff</a>


index.html
```diff
@@ -529,7 +529,7 @@ <h3>BitstringStatusListCredential</h3>
               </td>
             </tr>
             <tr>
-              <td>credentialSubject.ttl</td>
+              <td id="ttl">credentialSubject.ttl</td>
```
dlongley, line 532: I flagged this elsewhere -- but don't think `ttl` works as specified and we don't need it here. I won't object to it being added here, but we'll then just need to remove it later unless it can be shown that it works properly.

dlongley (approved):

index.html
msporny: Please raise an issue if you haven't done so already noting why TTL doesn't work.

msporny: Nevermind, I found it: https://github.com/w3c/vc-bitstring-status-list/issues/73#issuecomment-1607874866

iherman:
I will check the vocabulary generation file soonish (probably on the week-end), but one immediate remark that should be handled somehow.

- in contrast to the two other vocabularies, you chose to call the vocabulary file `index.yml` instead of `vocabulary.yml`
- the `yml2vocab` script's default name for the vocabulary file is `vocabulary.yml`
- the action script to include the vocabulary generation does not have an explicit name for the vocabulary file.
- the generation of the ttl and jsonld file is such that the 'base name' of the vocabulary file is used, i.e., if `index` is used then it will be `index.ttl` and index.jsonld`. The vocabulary template file, that refers to these formats, should reflect that.

All this means that the current setup will lead to a problem. I would suggest, to stay in line with the other vocabulary generation cases, to use `vocabulary.yml` file in instead of `index.yml`.

msporny:
@iherman wrote:
> All this means that the current setup will lead to a problem. 

Yes, I've been meaning to chat with you about the defaults. One of the downsides to having all the files generated say "vocabulary.html" is that browsing to the directory doesn't load the vocabulary file. On Github, you're just given a 404 (without directory contents).

I can change it back if it's a pain (simple rename), but I'm wondering if we should consider using whatever name the individual that wrote the file has chosen instead of forcing a naming convention on the people that use the `yml2vocab` tool. You are, of course, the final arbiter in that decision since you wrote the tool. :)

iherman:
> I can change it back if it's a pain (simple rename), but I'm wondering if we should consider using whatever name the individual that wrote the file has chosen instead of forcing a naming convention on the people that use the `yml2vocab` tool. You are, of course, the final arbiter in that decision since you wrote the tool. :)

I would propose to leave this as a `vocabulary.yml` for now. Any change would involve (1) make a (very simple) change in the yml2vocab package (2) update the package on npm (3) make all the changes in the DI and VCDM vocabularies as well, because the package taken from npm would not work for them anymore. Nothing complicated, but with the avalanche of PR-s and issues for CR, it is just a distraction for now. This is the typical issue that we can handle when we get to more peaceful times with CRs all around.

iherman:
@msporny while we are looking at those annoying minor issues... Is the namespace for this vocabulary (ie, `https://www.w3.org/ns/credentials/status-list#`) cast in concrete? The reason I am asking this is that we have now the following URL-s for the VC related vocabularies:

1. Credentials namespace: `https://www.w3.org/2018/credentials#`
2. DI namespace: `https://w3id.org/security#`
3. Bitstring status namespace: `https://www.w3.org/ns/credentials/status-list#`

Three different vocabularies with three totally different namespace URL-s: that is a real reason for confusion. We had some discussions before regarding (1) and (2), and there is a general agreement that, well, this is what we inherited, and it would lead to more problems changing those that it would solve. Ie, we are stuck with (1) and (2). But if so, wouldn't it be less confusing to the community if the bitstring status namespece was `https://www.w3.org/2018/status-list#`?

Using `/ns/credentials/...` would clearly be, in theory, nicer and cleaner. But isn't this the case where practical and user considerations should win over theoretical purity?

CC @dlongley

msporny:
@iherman wrote:
> I would propose to leave this as a `vocabulary.yml` for now.

Done in eb561c18f267652e743ea0a4fade6f322e4e5b4c.

> Is the namespace for this vocabulary (ie, https://www.w3.org/ns/credentials/status-list#) cast in concrete?

No, it's just a proposal for now. :)

It's based on our decision to do this in VCDM v2.0:

```
  "@context": [
    "https://www.w3.org/ns/credentials/v2",
    "https://www.w3.org/ns/credentials/examples/v2"
  ],
```

We didn't change the vocabulary URLs because we had deployed systems for VCDM v1.0 and v1.1 in production. The `StatusList2021` specification used `https://w3id.org/vc/status-list#` as it's vocabulary URL base. 

However, when the group decided to rename from `StatusList2021` we broke all backwards compatibility with all existing systems, and we had no systems that we knew of in production. So, this vocabulary URL is a bit of a green field exercise -- we are not held back by previous production deployments.

So, if `https://www.w3.org/ns/credentials/status-list#` is "nicer and cleaner", then we can do that for this particular vocabulary URL.

iherman:
> @iherman wrote:
> 
> > I would propose to leave this as a `vocabulary.yml` for now.
> 
> Done in [eb561c1](https://github.com/w3c/vc-bitstring-status-list/commit/eb561c18f267652e743ea0a4fade6f322e4e5b4c).
> 

Thanks. Let us come back to these when the CR storm settles :-)

> > Is the namespace for this vocabulary (ie, https://www.w3.org/ns/credentials/status-list#) cast in concrete?
> 
> No, it's just a proposal for now. :)
> 
[…]
> 
> > wouldn't it be less confusing to the community if the bitstring status namespece was https://www.w3.org/2018/status-list#?
> 
> Maybe? I wouldn't be opposed to that base URL. The only argument against it is that the status list vocabulary didn't exist in 2018. :)
> 
> So, if `https://www.w3.org/ns/credentials/status-list#` is "nicer and cleaner", then we can do that for this particular vocabulary URL.

This would be o.k. only if we decided to use `/ns/credentials/core#` or something like that for the credential vocabulary. That is a big no-no I believe it is even more clean and nice if these two vocabularies live side by side, so to say. 

Actually, in #81, I considered these terms to be simply added to the VCDM credential vocabulary... Would that be out of the question? Does it have a real value of keeping these two apart? (Just asking, I do not have strong feelings on that aspect.)

msporny:
@iherman wrote:
> Actually, in #81, I considered these terms to be simply added to the VCDM credential vocabulary... Would that be out of the question? Does it have a real value of keeping these two apart? (Just asking, I do not have strong feelings on that aspect.)

I'd be fine w/ this path as well, though, we might not want to get into a situation where the VC vocabulary becomes a massive vocabulary for all extensions to VCs as well. That would be a path towards centralization -- not everything important in VCs needs to go through the VCWG (or the VC vocabulary). 

It would be good for us to provide an example of how to extend the core data model, which is what this PR does. But, as I said above, I don't have strong feelings about this. I prefer that we provide a good example of an extension that we can point to in the future when others want to extend VCs in some way.

iherman:
>> Actually, in #81, I considered these terms to be simply added to the VCDM credential vocabulary... Would that be out of the question? Does it have a real value of keeping these two apart? (Just asking, I do not have strong feelings on that aspect.)
>
> I'd be fine w/ this path as well, though, we might not want to get into a situation where the VC vocabulary becomes a massive vocabulary for all extensions to VCs as well. That would be a path towards centralization -- not everything important in VCs needs to go through the VCWG (or the VC vocabulary).

Just for the sake of argument: moving these vocabulary items into the VCDM vocabulary does not mean that this is for _all_ extensions. It only means that we use it for extensions _defined by this Working Group_.

We already do something like that elsewhere: the few extra terms needed for the VC JSON Schema spec have been added to the VCDM vocabulary, instead of defining a separate vocabulary for those. I acknowledge that the comparison is not entirely ok, because this proposed vocabulary is a factor larger than the Schema spec's.

I would think this should be raised on the next WG call (I won't be there), and let the WG decide. I accept any decision done by the WG.

vocab/README.md
```diff
@@ -0,0 +1,11 @@
+# Generate RDFS vocabulary files
+
+The script in the directory generates RDFS vocabulary files in JSON and Turtle formats, plus a human readable HTML file containing the vocabulary in RDFa, based on a simple vocabulary definition in a YAML file. This is done using the [yml2vocab](https://github.com/w3c/yml2vocab); more details about the script can also be found in the [yml2vocab readme file](https://github.com/w3c/yml2vocab).
+
+The generation of the final files is done via a github action (see `/.github/workflows/generate-vocab-files.yml`).
+
+## Content of the directory
+
+- `README.md`: this file.
+- `index.yml`: the core vocabulary specification. _Any change on the vocabulary must be made by modifying this file;_ see the separate [description](https://github.com/w3c/yml2vocab) of the underlying YAML format.
```
iherman, line 10:
```suggestion
- `vocabulary.yml`: the core vocabulary specification. _Any change on the vocabulary must be made by modifying this file;_ see the separate [description](https://github.com/w3c/yml2vocab) of the underlying YAML format.
```

vocab/README.md
```diff
@@ -0,0 +1,11 @@
...
+- `template.html`: an HTML template file used by the script; it is the skeleton of the final HTML format based on [ReSpec](https://respec.org/docs/). If the file is modified, care should be taken not to change the core structure and the various, possibly empty, HTML elements with `@id` values. The script fills those elements with content when generating the `index.html` file (and removes any sections that remain empty after processing).
```
iherman, line 11:
```suggestion
- `template.html`: an HTML template file used by the script; it is the skeleton of the final HTML format based on [ReSpec](https://respec.org/docs/). If the file is modified, care should be taken not to change the core structure and the various, possibly empty, HTML elements with `@id` values. The script fills those elements with content when generating the `vocabulary.html` file (and removes any sections that remain empty after processing).
```

vocab/template.html
```diff
@@ -0,0 +1,226 @@
...
+          .replace(/href="https:\/\/www.w3.org\/ns\/credentials\/status-list#/g, 'href="#');
+        return retval;
+      }
+
+    </script>
+    <script class="remove">
+      var respecConfig = {
+        specStatus:  "base",
+        shortName:   "ns/credentials/status-list",
+        thisVersion: "https://www.w3.org/ns/credentials/status-list/",
```
iherman, line 26: Just flagging this, so that we should not forget: these entries may have to change if the decision is to choose a different namespace URL.

vocab/template.html
```diff
@@ -0,0 +1,226 @@
...
+      /* This is a bug in the base css setting at W3C. There is an open PR raised for it in the
+      relevant github repo; if and when that is merged, this statement may become
+      superfluous.
+      See https://github.com/w3c/tr-design/pull/336
+      */
+      figure svg {
+          max-width: 100%;
+          margin: auto;
+          height: auto;
+      }
```
iherman, line 108:
```suggestion

```

This can be removed now, the corresponding PR has been merged.

vocab/template.html
```diff
@@ -0,0 +1,226 @@
...
+      <p>
+In some cases, a local explanation is necessary to complement, or to replace,
+the definition found in an external specification. For instance, this is so when
+the term is needed to provide a consistent structure to the RDFS vocabulary,
+such as when the term defines a common supertype for class instances that are
+used as objects of specific properties, or when
+<a href="https://www.w3.org/TR/rdf12-concepts/#section-rdf-graph">RDF Graphs</a>
+are involved. For such cases, the extra definition is included in the current
+document (and the `rdfs:comment` property is used to include them in the RDFS
+representations).
```
iherman, line 167: Is this relevant? I.e., will this vocabulary ever deal with RDF Graphs? If not (which is probably the case) the whole section may be removed.

vocab/vocabulary.yml
```diff
@@ -0,0 +1,91 @@
+vocab:
+  - id: sl
+    value: https://www.w3.org/ns/credentials/status-list#
```
iherman, line 3: Same comment as above: this may have to change.

vocab/vocabulary.yml
```diff
@@ -0,0 +1,91 @@
+vocab:
+  - id: sl
```
iherman, line 2: Would it be user friendly to chose a different prefix, say, `slist`? The problem is that it is difficult to differentiate between `s1` and `sl` in many fonts, and this may be the source of stupid bugs...

vocab/vocabulary.yml
```diff
@@ -0,0 +1,91 @@
...
+prefix:
+  - id: cred
+    value: https://w3.org/2018/credentials#
+
+ontology:
+  - property: dc:title
+    value: Verifiable Credentials Bitstring Status List v1.0
+
+  - property: dc:description
+    value: RDFS [[RDF-SCHEMA]] vocabulary used by the Verifiable Credentials Bitstring Status List [[VC-BITSTRING-STATUS-LIST]]
```
iherman, line 14:
```suggestion
    value: RDFS [[RDF-SCHEMA]] vocabulary used by the [[[VC-BITSTRING-STATUS-LIST]]] [[VC-BITSTRING-STATUS-LIST]]
```

vocab/vocabulary.yml
iherman: (I made the mistake reviewing this file in github: I was wondering why on earth do we use the prefix `s1`, ie, "sh one"...)

iherman (changes requested):

TallTed (approved):

vocab/template.html
msporny: Correct, this vocabulary will probably never deal w/ RDF Graphs. We can remove the statement about RDF Graphs.

vocab/template.html
```diff
@@ -0,0 +1,226 @@
...
+specification.
+      </p>
+      <p>
+In some cases, a local explanation is necessary to complement, or to replace,
+the definition found in an external specification. For instance, this is so when
+the term is needed to provide a consistent structure to the RDFS vocabulary,
+such as when the term defines a common supertype for class instances that are
+used as objects of specific properties, or when
+<a href="https://www.w3.org/TR/rdf12-concepts/#section-rdf-graph">RDF Graphs</a>
+are involved. For such cases, the extra definition is included in the current
```
msporny, line 165:
```suggestion
used as objects of specific properties. For such cases, 
the extra definition is included in the current
```

vocab/vocabulary.yml
msporny: Now for the important question... if not `sl`, then what? :P 

Would `sv` be better? For "Status Vocabulary"? and then perhaps shorten the URL to: `https://www.w3.org/ns/credentials/status#` ?

vocab/vocabulary.yml
msporny: What about `https://www.w3.org/ns/credentials/status#` ?

vocab/vocabulary.yml
iherman: I think `sv` is really better.

As for the URL, for the reasons I started this discussions with, my personal preference is to put it into /2018/, alongside the VCDM vocabulary.

vocab/vocabulary.yml
iherman: See my answer below

vocab/vocabulary.yml
iherman: (but shorten it to `.../status` is indeed a good idea imho.)

iherman: Just to be precise, because I made a mistake in the original comment, what about `https://www.w3.org/2018/credentials/status#`

(Noting that we will have to update the `.htaccess` file in `/2018/credentials/`)

iherman:
Just flagging this: before merging (if there is an approval to do so) some name changes that happened in other PR-s should be added here. See https://github.com/w3c/vc-bitstring-status-list/commit/35860acfc10c8ed7db692e1ef52d43655d770176 (but there may be more).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-10-vcwg#section3-1) on 2024-01-10

- no resolutions were taken

#### 3.1. Rename "status" to "message" for `statusPurpose` feature. (pr vc-bitstring-status-list#100)

_See github pull request [vc-bitstring-status-list#100](https://github.com/w3c/vc-bitstring-status-list/pull/100)._




**Manu Sporny:** we had 3 status purpose values: revocation, suspension and status.  
… we have changed status to message.  
… we have pre-pended the word status to some of the property names.  

_See github pull request [vc-bitstring-status-list#105](https://github.com/w3c/vc-bitstring-status-list/pull/105)._




</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-10-vcwg#section3-8) on 2024-01-10

- no resolutions were taken

#### 3.8. Add formal vocabulary definitions. (pr vc-bitstring-status-list#105)

_See github pull request [vc-bitstring-status-list#105](https://github.com/w3c/vc-bitstring-status-list/pull/105)._




**Manu Sporny:** what is the vocabulary URL for the status list?  
… several proposals for this.  

> *Manu Sporny:* This could be one of the vocabulary URL prefixes: `https://www.w3.org/ns/credentials/status-list#`.

> *Manu Sporny:* This could be one of the vocabulary URL prefixes: `https://www.w3.org/2018/credentials/status#`.

**Manu Sporny:** we need to pick one and go with it.  

> *Manu Sporny:* This could be one of the vocabulary URL prefixes: `https://www.w3.org/ns/credentials/status#`

**Manu Sporny:** but does not align with 2018 existing terms.  
… should there be a date in the terms or not?  

**Brent Zundel:** do we want to have a poll on the name?  

**POLL:** What should the status list vocabulary URL be Choice A) `https://www.w3.org/ns/credentials/status#` and Choice B) `https://www.w3.org/2018/credentials/status#` ?  

> *Ivan Herman:* B.

> *Manu Sporny:* A.

> *Andres Uribe:* A.

> *Brent Zundel:* A.

> *Phillip Long:* A.

> *David Chadwick:* B.

> *Paul Dietrich:* A.

> *Joe Andrieu:* B.

> *Kevin-Dean-GS1:* A.

**David Chadwick:** The reason I wanted the date is if you want to change a definition, if you have date in there, change in definition could have a new date. I see how you can append/remove, but changing them is not possible.  

> *Ted Thibodeau Jr.:* I don't fully grok the pros and cons. A `ns` vs B `2018` (which is when the original VCWG started).

**Ivan Herman:** There might be a version in the vocabulary, but URL does not change.  

---


</details>

David-Chadwick (approved):

vocab/vocabulary.yml
iherman: Per [WG meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-10-vcwg#section3-8) this issue is getting closed. But I think using "sv" instead of "sl" would still be better, and that means that URL would be `/ns/credentials/status#`.

Once this is merged, the `.htaccess` in `/ns/credentials` will need some magic...

vocab/vocabulary.yml
iherman: Per [WG meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-10-vcwg#section3-8) `/ns/credentials/status#` should be o.k.

msporny: Fixed in 636451dd7676b2922905dc919695054fec0af6e6

vocab/vocabulary.yml
msporny: Updated in 636451dd7676b2922905dc919695054fec0af6e6

vocab/template.html
msporny: Fixed in aa07c21ffebac1c6456a98c0e4352be392bacd4a.

vocab/template.html
msporny: Mentions of RDF Graphs have been removed in the PR.

msporny:

Issue #64, opened 2023-05-15
Add section on expected server HTTP status codes


msporny:
It has been suggested that the specification should articulate what HTTP status codes are sent by servers, and expected by clients, when certain operations, such as dereferencing occur. This issue has been created to track that concern.

OR13:
we should also comment on accept / content type headers as well.

Sakurann:
TPAC VC WG: normative statements. idea is to learn from DI example.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-15-vcwg#section1-4) on 2023-09-15

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.4. Add section on expected server HTTP status codes (issue vc-status-list-2021#64)

_See github issue [vc-status-list-2021#64](https://github.com/w3c/vc-status-list-2021/issues/64)._




**Manu Sporny:** need approval from the group to talk about potential http response status codes ... request by mike jones.  
… if you do things with status lists with HTTP, what should you expect ...  
… from HTTP status codes ...  
… if you operate over http these are the things that you can expect to see from servers ....  
… looking to see if anyone that would oppose to adding a section on operating over http and expecting http error codes.  

**Brent Zundel:** any opposition as that for a path forward?  

**Kristina Yasuda:** sgtm.  
… when this was raised we deliberately didn't do right away ... what has changed? why is it ok to do this now?  

**Joe Andrieu:** just wanted to clarify, the intention is to not use non-normative language, right?  

**Manu Sporny:** what has changed, kristina, is that we figured out a way to do it without raising objections, a design pattern in the status list that people din't reject.  

> *Kristina Yasuda:* it needs to be normative.

**Manu Sporny:** joe, if you are using HTTP these things MUST be done.  

> *Joe Andrieu:* +1 to normative. just wanted to understand the scope of the proposal.

> *Brent Zundel:* prefer normative, since we've found a way to do that.

**Manu Sporny:** mike, wdyt, would you more normative language around this?  

**mike:** yeah.  

**Manu Sporny:** as long as no one opposed, we'd use normative language.  
… i'll take that.  

</details>

msporny:
PR #111 has been raised to address this issue. This issue will be closed once PR #111 has been merged.

msporny:
PR #111 has been merged, closing.

PR #111, opened 2023-12-29
Add section on Processing Errors.

msporny:
This PR attempts to address issue #64 by adding a section about processing errors and including how one can respond over HTTP if the APIs are exposed over HTTP.

### DIFF
--- a/index.html
+++ b/index.html
@@ -718,12 +718,15 @@ in the `credentialStatus` entry in the
         </li>
         <li>
 Dereference the `statusListCredential` URL, and ensure that all
-proofs verify successfully. If the dereference fails, or if any of the proof
-verifications fail, return a validation error.
+proofs verify successfully. If the dereference fails, raise a
+<a href="#STATUS_RETRIEVAL_ERROR">STATUS_RETRIEVAL_ERROR</a>. If any of the
+proof verifications fail, raise a
+<a href="#STATUS_VALIDATION_ERROR">STATUS_VALIDATION_ERROR</a>.
         </li>
         <li>
-Verify that the |status purpose| matches the
-`statusPurpose` value in the |statusListCredential|.
+Verify that the |status purpose| is equal to the
+`statusPurpose` value in the |statusListCredential|. If the values are not
+equal, raise a <a href="#STATUS_VALIDATION_ERROR">STATUS_VALIDATION_ERROR</a>.
         </li>
         <li>
 Let |compressed bitstring| be the value of the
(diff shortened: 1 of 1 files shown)


TallTed (approved):

index.html
```diff
@@ -718,12 +718,15 @@ <h3>Validate Algorithm</h3>
         </li>
         <li>
 Dereference the `statusListCredential` URL, and ensure that all
-proofs verify successfully. If the dereference fails, or if any of the proof
-verifications fail, return a validation error.
+proofs verify successfully. If the dereference fails, raise a
+<a href="#STATUS_RETRIEVAL_ERROR">STATUS_RETRIEVAL_ERROR</a>. If any of the
+proof verifications fail, raise a
+<a href="#STATUS_VALIDATION_ERROR">STATUS_VALIDATION_ERROR</a>.
```
dlongley, line 724: Maybe this should be `STATUS_VERIFICATION_ERROR` instead? Same for each other use of this error; it seems they may just be about verifying the data is authentic and proper vs. business rules around accepting the status information.

dlongley (approved):

index.html
TallTed: I think the status fails to validate.

What the verifier chooses to do about that failure is then a business rules question.

dlongley: Ok, I can support that mental model.

David-Chadwick: I would just like to check that we have the following alternative processes covered:
1. the verifier asks the issuer for a status list at some previous date.time (e.g. last week)
2. the issuer responds with a status retrieval error because such a list is not available
3. the verifier then either terminates or asks for the current status list
Alternatively 
1a. the verifier asks the issuer for a status list at some previous date.time (e.g. last week)
3a the issuer  ignores the date.time on the query, and returns the current status list.

TallTed: > I would just like to check that we have the following alternative processes covered:
>
> 1. the verifier asks the issuer for a status list at some previous date.time (e.g. last week)
> 1. the issuer responds with a status retrieval error because such a list is not available
> 1. the verifier then either terminates or asks for the current status list

Step 3 feels very much like business logic to me.

> Alternatively
> 1a. the verifier asks the issuer for a status list at some previous date.time (e.g. last week)
> 2a. the issuer ignores the date.time on the query, and returns the current status list.

I do not think (2a) is viable *unless* the status list VC includes its validity datetime range, which must be included whether or not the issuer supports historical status lists ... and I think there ought to be an advisory message along those lines ("You asked for a historical status list. I don't provide those. Here's my current status list.") to ensure that the verifier knows they're not looking at the historical snapshot they requested.

brianorwhatever (approved):

KDean-GS1 (approved):

index.html
msporny: @dlongley your concerns have been fixed in 6e6d6b963e8f6c62fb5c1a2f792bbefc6c4d0b67.

@David-Chadwick I'll address your concern in PR #112 since it has to do w/ queries for historical information.

msporny:

Issue #87, opened 2023-09-15
Should validation algorithm check size of bitstring?


andresuribe87:
The section in https://w3c.github.io/vc-status-list-2021/#validate-algorithm does not mention checking the minimum bitstring size.

Suggestion is that validation should fail when the size of the bitstring is below the minimum of 16KB.

msporny:
PR #115 has been raised to address this issue. This issue will be closed once PR #115 has been merged.

msporny:
PR #115 has been merged, closing.

PR #109, opened 2023-12-29
Update BitstringStatusList context to align with latest spec.

msporny:
This PR is an attempt at addressing issue #70, which updates the context to align with the latest specification. 

It preserves the old 2021 status list since that's where the w3id.org redirect goes currently, and we believe some organizations have deployed 2021 status list into potentially production environments. We need to preserve that file for now in order to not break implementations that aren't (mistakenly) caching their JSON-LD Contexts until we can move it back to the CCG or host it elsewhere (or decide to continue hosting it here).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/109.html" title="Last updated on Jan 13, 2024, 4:38 PM UTC (33d25b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/109/887193d...33d25b7.html" title="Last updated on Jan 13, 2024, 4:38 PM UTC (33d25b7)">Diff</a>

PR #115, opened 2023-12-29
Raise error when status list size does not meet minimum threshold.

msporny:
This PR is an attempt to address issue #87 by raising an error when status list size does not meet minimum threshold.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/115.html" title="Last updated on Jan 13, 2024, 7:42 PM UTC (601246f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/115/26aaaf5...601246f.html" title="Last updated on Jan 13, 2024, 7:42 PM UTC (601246f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,15 +245,16 @@ One of the benefits of using a bitstring is that it is a highly compressible
 data format since, in the average case, large numbers of credentials will
 remain unrevoked. This will ensure long sections of bits that are the same
 value and thus highly compressible using run-length compression techniques
-such as GZIP [[RFC1952]]. The default bitstring size is 16KB (131,072 entries),
-and when only a handful of <a>verifiable credentials</a> are revoked, the
-compressed bitstring size is reduced down to a few hundred bytes.
+such as GZIP [[RFC1952]]. The default status list size is 131,072 entries,
+equivalent to 16KB of single bit values. When only a handful of <a>verifiable
+credentials</a> are revoked, the compressed bitstring size is reduced down to a
+few hundred bytes.
         </p>
 
         <p>
 Another benefit of using a bitstring is that it enables large numbers of
-<a>verifiable credential</a> statuses to be placed in the same  list.
-This specification utilizes a minimum bitstring length of 131,072 (16KB). This
+<a>verifiable credential</a> statuses to be placed in the same list.
+This specification utilizes a minimum list length of 131,072. This
 population size ensures an adequate amount of herd  privacy in the average case.
 If better herd privacy is required, the bitstring can be made to be larger.
         </p>
@@ -779,6 +780,11 @@ Generate a |revocation bitstring| by passing
 <a href="#bitstring-expansion-algorithm">Bitstring Expansion Algorithm</a>.
         </li>
         <li>
+If the length of the |revocation bitstring| divided by
+<a href="#size">`size`</a> is less than 131,072, raise a
+<a href="#STATUS_LIST_LENGTH_ERROR">STATUS_LIST_LENGTH_ERROR</a>.
+        </li>
+        <li>
 Let |status| be the value in the |bitstring| at the position indicated
 by the |credentialIndex| multiplied by the |size|. If the |credentialIndex|
 multiplied by the |size| is a value outside of the range of the |bitstring|, a
@@ -929,7 +935,9 @@ The `detail` value SHOULD provide a longer human-readable string for the error.
       </ul>
 
       <dl>
-        <dt id="STATUS_RETRIEVAL_ERROR">STATUS_RETRIEVAL_ERROR (-128)</dt>
+        <dt id="STATUS_RETRIEVAL_ERROR">
+STATUS_RETRIEVAL_ERROR (-128)
+        </dt>
         <dd>
 Retrieval of the status list failed. See Section
 <a href="#validate-algorithm"></a>.
@@ -938,6 +946,13 @@ Retrieval of the status list failed. See Section
         <dd>
 Validation of the status entry failed. See Section
 <a href="#validate-algorithm"></a>.
+        </dd>
+        <dt id="STATUS_LIST_LENGTH_ERROR">
+STATUS_LIST_LENGTH_ERROR (-130)
+        </dt>
+        <dd>
+The status list does not meet the minimum herd privacy length requirements.
+See Section <a href="#validate-algorithm"></a>.
         </dd>
       </dl>
     </section>
--- a/index.html
+++ b/index.html
@@ -246,17 +246,17 @@ data format since, in the average case, large numbers of credentials will
 remain unrevoked. This will ensure long sections of bits that are the same
 value and thus highly compressible using run-length compression techniques
 such as GZIP [[RFC1952]]. The default status list size is 131,072 entries,
-equivalent to 16KB of single bit values. When only a handful of <a>verifiable
-credentials</a> are revoked, the compressed bitstring size is reduced down to a
-few hundred bytes.
+equivalent to 16&nbsp;KB of single bit values. When only a handful of
+<a>verifiable credentials</a> are revoked, GZIP compresses the bitstring
+to a few hundred bytes.
         </p>
 
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -244,15 +244,16 @@ <h3>Conceptual Framework</h3>
 data format since, in the average case, large numbers of credentials will
 remain unrevoked. This will ensure long sections of bits that are the same
 value and thus highly compressible using run-length compression techniques
-such as GZIP [[RFC1952]]. The default bitstring size is 16KB (131,072 entries),
-and when only a handful of <a>verifiable credentials</a> are revoked, the
-compressed bitstring size is reduced down to a few hundred bytes.
+such as GZIP [[RFC1952]]. The default status list size is 131,072 entries,
+equivalent to 16KB of single bit values. When only a handful of <a>verifiable
+credentials</a> are revoked, the compressed bitstring size is reduced down to a
+few hundred bytes.
```
TallTed, line 250:
```suggestion
equivalent to 16&nbsp;KB of single bit values. When only a handful of
<a>verifiable credentials</a> are revoked, GZIP compresses the bitstring
to a few hundred bytes.
```

index.html
```diff
@@ -244,15 +244,16 @@ <h3>Conceptual Framework</h3>
...
         </p>
 
         <p>
 Another benefit of using a bitstring is that it enables large numbers of
-<a>verifiable credential</a> statuses to be placed in the same  list.
-This specification utilizes a minimum bitstring length of 131,072 (16KB). This
+<a>verifiable credential</a> statuses to be placed in the same list.
+This specification utilizes a minimum list length of 131,072. This
 population size ensures an adequate amount of herd  privacy in the average case.
 If better herd privacy is required, the bitstring can be made to be larger.
```
TallTed, line 258:
```suggestion
This specification uses a minimum list length of 131,072. This
size ensures an adequate amount of herd privacy in the average case.
If better herd privacy is required, the bitstring can be made larger.
```

index.html
```diff
@@ -744,6 +745,11 @@ <h3>Validate Algorithm</h3>
 <a href="#bitstring-expansion-algorithm">Bitstring Expansion Algorithm</a>.
         </li>
         <li>
+If the length of the |revocation bitstring| divided by
+<a href="#size">`size`</a> is less than 131,072, raise a
```
TallTed, line 749:
```suggestion
<a href="#statusSize">`statusSize`</a> is less than 131,072, raise a
```

index.html
```diff
@@ -852,16 +858,27 @@ <h3>Processing Errors</h3>
...
         <dd>
 Validation of the status entry failed. See Section
 <a href="#validate-algorithm"></a>.
         </dd>
+        <dt id="STATUS_LIST_LENGTH_ERROR">
+STATUS_LIST_LENGTH_ERROR (-130)
+        </dt>
+        <dd>
+The status list does not meet the minimum herd privacy length requirements.
+See Section <a href="#validate-algorithm"></a>.
```
TallTed, line 880:
```suggestion
The status list length does not satisfy the minimum length required for
herd privacy. See Section <a href="#validate-algorithm"></a>.
```

TallTed (changes requested):

index.html
msporny: I'm going to hold off on this change until the change happens in #109.

index.html
```diff
@@ -244,17 +244,18 @@ <h3>Conceptual Framework</h3>
...
 
         <p>
 Another benefit of using a bitstring is that it enables large numbers of
-<a>verifiable credential</a> statuses to be placed in the same  list.
-This specification utilizes a minimum bitstring length of 131,072 (16KB). This
-population size ensures an adequate amount of herd  privacy in the average case.
-If better herd privacy is required, the bitstring can be made to be larger.
+<a>verifiable credential</a> statuses to be placed in the same list.
+This specification uses a minimum list length of 131,072. This
+size ensures an adequate amount of herd privacy in the average case.
```
dlongley, line 257: The suggestion has been to replace "herd" with "group" everywhere as friendlier terminology.

```suggestion
size ensures an adequate amount of group privacy in the average case.
```

index.html
dlongley, line 749: I'm a little apprehensive about us forcing implementations to raise this error when perhaps there are legitimate use cases for a smaller size. We started with a default size that we thought was good for the average case, but this PR is now forcing that to be a hard minimum. I think we should be more flexible.

```suggestion
<a href="#size">`size`</a> is less than 131,072, implementations
SHOULD raise a
```

David-Chadwick: I am not sure how SHOULD would actually work in practice. Say a general purpose wallet implements the SHOULD as a MUST then it will always raise an error on short lists. But if this wallet implements the SHOULD as a MAY, then it will never raise an error on short lists and will always accept them. I would not advocate asking the user whether a short list should be accepted or not. Therefore from an implementation perspective I think that MUST should be the standard definition and verifiers should always raise an error on short lists. If a particular eco-system wants to implement short lists then its issuers and wallets and verifiers can do this without affecting the global eco-system.

David-Chadwick (approved):
Without changing MUST to SHOULD

index.html
msporny: Yes, I was thinking the same thing while processing these PRs. I was going to raise an entirely separate PR to do the "herd" -> "group" renaming.

msporny: Change raised in PR #121.

index.html
msporny: Yes, I'm somewhat concerned as well... but couldn't think of a use case where a smaller size was warranted.

If the population size is smaller, the 16KB array compresses down to a very small number (a few hundred to thousand bytes in size). So, the argument against needing a smaller size probably doesn't hold... and that's the only argument I can see for allowing a smaller list size.

mkhraisha (approved):

index.html
dlongley: @David-Chadwick,

> I am not sure how SHOULD would actually work in practice. Say a general purpose wallet implements the SHOULD as a MUST then it will always raise an error on short lists. But if this wallet implements the SHOULD as a MAY, then it will never raise an error on short lists and will always accept them. I would not advocate asking the user whether a short list should be accepted or not.

But I think this is precisely how a SHOULD would work in practice. A very generic wallet (or verifier) would implement it "as a MUST" as you say and it would not accept anything outside of those bounds.

> Therefore from an implementation perspective I think that MUST should be the standard definition and verifiers should always raise an error on short lists. If a particular eco-system wants to implement short lists then its issuers and wallets and verifiers can do this without affecting the global eco-system.

This would cause those implementations (in that particular ecosystem) to be in violation of the standard, which is the undesirable outcome that the SHOULD is intended to avoid.

Perhaps one way to resolve this is to make it a MUST with an "unless clause" that says that there's an agreed upon smaller size for particular credentials within a subecosystem or something along those lines.

David-Chadwick: I do not think it is an undesirable outcome if an isolated eco-system chooses to ignore one or more parts of the standard and to implement its own features. Every isolated eco-system is free to implement what it wants to. We should be concerned about the global interworking system, which is why MUST is the preferred statement in the standard.

TallTed: We SHOULD also be somewhat concerned with such an "isolated eco-system" eventually joining "the global interworking system", or even needing to be blended with a *second* "isolated eco-system". SHOULD is a far better future-proof than MUST, even though it allows two such "isolated eco-system" to imperfectly interoperate because one raises the error and the other does not.

msporny: This change is now applied in #109.

msporny: Fixed in 601246f3a8c22ba8647f1b2aeadca805c08a6897 in a way that achieves what I believe @dlongley, @David-Chadwick, and @TallTed wanted to see.

The minimum number of entries is 131,072 unless a different one is specified by a particular specification and provided as input by the application (presumably through validation rules). The algorithm MUST throw an error if the length of the list is below the acceptable threshold established.

msporny:

PR #122, opened 2023-12-30
Add examples for multiple status lists and multiple entries in a single status list

msporny:
This PR is an attempt at addressing [some comments](https://github.com/w3c/vc-bitstring-status-list/issues/73#issuecomment-1872490184) in issue #73 by adding examples for multiple status lists and multiple entries in a single status list.

Note: This PR contains a normative change to support the last example with multiple status types stored in a single list. It also has an at-risk marker for the feature as the more I wrote about the feature, the less of a benefit there seemed to be. The strongest benefit seems to be a privacy argument (if you mix in multiple statuses in a single list, and the bits for each status don't sit beside one another, it's hard to tell which type of status bit is being flipped to an outside observer -- which improves the privacy characteristics of the list).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/122.html" title="Last updated on Jan 13, 2024, 9:19 PM UTC (4a51b3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/122/c53823f...4a51b3b.html" title="Last updated on Jan 13, 2024, 9:19 PM UTC (4a51b3b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1312,6 +1312,11 @@ data-vc-vm='https://example.edu/issuers/565049/keys/1'>
       <section>
         <h2>Multiple Status Lists in One Verifiable Credential</h2>
 
+        <p>
+This specification enables an <a>issuer</a> to associate multiple status lists
+with a single <a>verifiable credential</a>.
+        </p>
+
         <pre class="example nohighlight vc"
              title="Associating multiple status lists with a single Verifiable Credential"
              data-vc-vm='https://example.edu/issuers/565049/keys/1'>
@@ -1349,6 +1354,33 @@ data-vc-vm='https://example.edu/issuers/565049/keys/1'>
 
       <section>
         <h2>Multiple Status Entries in a Single List</h2>
+
+        <p>
+It is possible for a single status list to contain multiple types of status
+purposes. Doing so can make the retrieval of a list slightly more efficient
+than fetching multiple status lists.
+        </p>
+
+        <p class="issue atrisk" title="Efficiency argument is weak">
+The "space efficiency" argument for this feature is weak. Fetching one list
+with two types of status entries must, presumably, be twice as long to ensure
+proper privacy protections. One privacy benefit of doing so is that bit flips
+cannot be known to be associated with a particular status unless one is also in
+control of the VC that the status is about. Therefore, mixing "revocation" and
+"suspension" in a single list that is twice as large has positive privacy
+implications.<br><br>
+The "retrieval efficiency" argument is also weak. Performing two HTTP retrievals
+instead of one is probably not significant. Performing upwards of five to six,
+on a list that is five to six times larger, might result in fairly meager
+savings over modern versions of HTTP that bundle requests over a single channel
+(such as HTTP/2 or HTTP/3). The requests themselves would be a handful of bytes
+that are saved with no significant improvement in retrieval speed.<br><br>
+The Working Group is looking for feedback from implementers and is considering
+striking this feature during the Candidate Recommendation period since it would
+simplify the specification for implementations to not have to support sets of
+`proofPurpose` values in the status list credentials (again, a meager savings in
+complexity).
+        </p>
 
         <pre class="example nohighlight vc"
              title="Associating multiple status entries in a single status list"
--- a/index.html
+++ b/index.html
@@ -512,9 +512,9 @@ status list, MUST be `BitstringStatusList`.
             <tr>
               <td id="statusPurpose">credentialSubject.statusPurpose</td>
               <td>
-The purpose of the status entry MUST be one or more strings. While the value of
-each string is arbitrary, the following values MUST be used for their intended
-purpose:
+The value of the purpose property of the status entry, `statusPurpose`, MUST be
+one or more strings. While the value of each string is arbitrary, the following
+values MUST be used for their intended purpose:
                 <table class="simple">
                   <thead>
                     <tr>
@@ -1379,7 +1379,7 @@ that are saved with no significant improvement in retrieval speed.<br><br>
 The Working Group is looking for feedback from implementers and is considering
 striking this feature during the Candidate Recommendation period since it would
 simplify the specification for implementations to not have to support sets of
-`proofPurpose` values in the status list credentials (again, a meager savings in
+`statusPurpose` values in the status list credentials (again, a meager savings in
 complexity).
         </p>
 
--- a/index.html
+++ b/index.html
@@ -1256,6 +1256,7 @@ Write i18n considerations.
     <h2>Appendix</h2>
     <section>
       <h2>Examples</h2>
+
       <section>
         <h2>Revocable Verifiable Credential</h2>
 <pre class="example nohighlight vc" title="A Revocable Verifiable Credential"
@@ -1284,6 +1285,7 @@ data-vc-vm='https://example.edu/issuers/565049/keys/1'>
 </pre>
 
       </section>
+
       <section>
         <h2>Status List Verifiable Credential</h2>
 <pre class="example nohighlight vc" title="A Status List Verifiable Credential"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+The "retrieval efficiency" argument is also weak. Performing two HTTP retrievals
+instead of one is probably not significant. Performing upwards of five to six,
+on a list that is five to six times larger, might result in fairly meager
+savings over modern versions of HTTP that bundle requests over a single channel
+(such as HTTP/2 or HTTP/3). The requests themselves would be a handful of bytes
+that are saved with no significant improvement in retrieval speed.<br><br>
+The Working Group is looking for feedback from implementers and is considering
+striking this feature during the Candidate Recommendation period since it would
+simplify the specification for implementations to not have to support sets of
+`proofPurpose` values in the status list credentials (again, a meager savings in
```
dlongley, line 1151: I think this was a typo?:

```suggestion
`statusPurpose` values in the status list credentials (again, a meager savings in
```

dlongley (approved):
Approving, but generally not a fan of the approach, I think it adds unnecessary complexity.

David-Chadwick (commented):
Why dont we remove the multiple entries feature before CR in the interests of simplification?

index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+complexity).
```
TallTed, line 1152: `a meager savings in complexity` seems to go against the rest of this note, which discusses meager savings in time and/or data size, gained at the expense of more complex implementations...
```suggestion
`proofPurpose` values in the status list credentials (again, a meager savings in
complexity).
```

TallTed (changes requested):

index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+instead of one is probably not significant. Performing upwards of five to six,
+on a list that is five to six times larger, might result in fairly meager
+savings over modern versions of HTTP that bundle requests over a single channel
+(such as HTTP/2 or HTTP/3). The requests themselves would be a handful of bytes
+that are saved with no significant improvement in retrieval speed.<br><br>
+The Working Group is looking for feedback from implementers and is considering
+striking this feature during the Candidate Recommendation period since it would
+simplify the specification for implementations to not have to support sets of
+`statusPurpose` values in the status list credentials (again, a meager savings in
+complexity).
```
msporny, line 1152:
```suggestion
`statusPurpose` values in the status list credentials (again, a meager savings in
space efficiency vs. retrieval efficiency).
```

brianorwhatever (approved):

KDean-GS1 (commented):
I see no reason for having multiple status lists; I'm hard-pressed to imagine a scenario where a verifier would be interested in "this" status but not "that" status. I'm sure such a scenario exists, but it's probably in a very small minority.

I'm neutral on this change.

longpd (commented):
I'm also not persuaded by this requested change. The security benefits and time savings seem minor. As KDean-GS1 noted "I'm hard-pressed to imagine a scenario where a verifier would be interested in "this" status but not "that" status.". Not sure the complexity is worth it - though will remain neutral and not object.

msporny:
@KDean-GS1 wrote:
> I see no reason for having multiple status lists; I'm hard-pressed to imagine a scenario where a verifier would be interested in "this" status but not "that" status. I'm sure such a scenario exists, but it's probably in a very small minority.

Here's a fairly significant set of use cases that care about revocation, but probably don't care about suspension: using any government issued identification to verify:

* birthday
* citizenship
* current address
* height
* weight
* biometric photo

IOW, just because the document is suspended (but not revoked) probably doesn't affect the claims above. This can be true for a variety of other document types as well.

I'm offering this as a counter-point, not arguing for any particular direction on this PR.

KDean-GS1:
> @KDean-GS1 wrote:
> 
> > I see no reason for having multiple status lists; I'm hard-pressed to imagine a scenario where a verifier would be interested in "this" status but not "that" status. I'm sure such a scenario exists, but it's probably in a very small minority.
> 
> Here's a fairly significant set of use cases that care about revocation, but probably don't care about suspension: using any government issued identification card to verify:
> 
> * name
> * birthday
> * gender
> * country of domicile
> * current address
> * height
> * weight
> * biometric photo
> * sight restrictions
> 
> IOW, just because the document is suspended (but not revoked) probably doesn't affect the claims above. This can be true for a variety of other document types as well.
> 
> I'm offering this as a counter-point, not arguing for any particular direction on this PR.

I'm not sold on that. If the government has suspended an identification credential, it could be because they are investigating the holder for some fraudulent declaration. For example, Toronto, Ontario has a law requiring that short-term renters (e.g., via Airbnb) reside at the address they're renting out, and proving it via a driver's license. This has led to families with multiple properties having individuals within the family each declare a property as their primary residence and reflecting it on their driver's license. If the government believes that the address declarations are fraudulent, they could suspend the credentials pending proper proof of residency.

Basically, unless specific use case rules say otherwise, any credential that with a status other than "active" should be treated at least as suspect. I think the process should be as follows: check the status, and once the status is known, whatever it might be, then and only then apply the validation rules.

TallTed:
> I think the process should be as follows: check the status, and once the status is known, whatever it might be, then and only then apply the validation rules.

That would be dictating business logic, which is beyond our purview.

I think we *can* legitimately say that Verifiers SHOULD check the status, but we _cannot_ say that they MUST nor SHOULD do anything based on the result.

KDean-GS1:
> That would be dictating business logic, which is beyond our purview.

Agreed. This is more about a generic approach; I can't imagine building a library that would selectively ignore any status instead of "get everything and pass it to the caller to figure out".

> I think we _can_ legitimately say that Verifiers SHOULD check the status, but we _cannot_ say that they MUST nor SHOULD do anything based on the result.

Definitely.

index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+        <h2>Multiple Status Lists in One Verifiable Credential</h2>
+
+        <p>
+This specification enables an <a>issuer</a> to associate multiple status lists
+with a single <a>verifiable credential</a>.
+        </p>
+
+        <pre class="example nohighlight vc"
+             title="Associating multiple status lists with a single Verifiable Credential"
+             data-vc-vm='https://example.edu/issuers/565049/keys/1'>
```
TallTed, line 1092: Is there a reason that `'` is used to wrap `data-vc-vm` values (lines 1092 and 1157), instead of `"` as on all other attribute values? It just seems odd to differ like this...

index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+}
+        </pre>
+      </section>
+
+      <section>
+        <h2>Multiple Status Entries in a Single List</h2>
+
+        <p>
+It is possible for a single status list to contain multiple types of status
+purposes. Doing so can make the retrieval of a list slightly more efficient
```
TallTed, line 1130:
```suggestion
purpose. Doing so can make the retrieval of a list slightly more efficient
```

index.html
```diff
@@ -1075,6 +1078,115 @@ <h2>Status List Verifiable Credential</h2>
...
+instead of one is probably not significant. Performing upwards of five to six,
+on a list that is five to six times larger, might result in fairly meager
+savings over modern versions of HTTP that bundle requests over a single channel
+(such as HTTP/2 or HTTP/3). The requests themselves would be a handful of bytes
+that are saved with no significant improvement in retrieval speed.<br><br>
+The Working Group is looking for feedback from implementers and is considering
+striking this feature during the Candidate Recommendation period since it would
+simplify the specification for implementations to not have to support sets of
+`statusPurpose` values in the status list credentials (again, a meager savings in
+space efficiency vs. retrieval efficiency).
```
TallTed, line 1152:
```suggestion
The "space efficiency" argument for this feature is weak. One list with two types
of status entries must, presumably, be twice as long as a list with one type of
status entries, to ensure proper privacy protections. One privacy benefit of
doing so is that bit flips cannot be known to be associated with a particular
status unless one is also in control of the VC that the status is about.
Therefore, mixing "revocation" and "suspension" in a single list that is twice
as large has positive privacy implications.<br><br>
The "retrieval efficiency" argument is also weak. Performing two HTTP retrievals
instead of one is probably not significant. Performing upwards of five or six,
on a list that is five or six times larger, might result in fairly meager
savings over modern versions of HTTP that bundle requests over a single channel
(such as HTTP/2 or HTTP/3). The requests themselves would save a handful of
bytes with no significant improvement in retrieval speed.<br><br>
The Working Group is looking for feedback from implementers and is considering
striking this feature during the Candidate Recommendation period, since it would
simplify the specification for implementations to not have to support sets of
`statusPurpose` values in the status list credentials (again, a meager savings
in space efficiency at a small cost to retrieval efficiency).
```

TallTed (changes requested):
minor

index.html
msporny: No, no reason, just a typo.

msporny: Fixed in 8d234dfed27adc8c84ad49ce1b7da28c41ca39ab

msporny:

Issue #150, opened 2024-03-06
Describe decoy entries as a privacy mitigation


npdoty:
Could dummy entries in the status list help to protect against an attacker who just regularly scans the list to try to reidentify users or their status?

Spec should document resistance to statistical analyses.

msporny:
Yes, and it would be worth saying something about that in the specification. We went into what one can do to avoid statistical attacks against group privacy during the presentation on bitstring status list yesterday:

https://meet.w3c-ccg.org/archives/w3c-ccg-weekly-2024-02-20.mp4

We should speak to decoy values and how one can avoid statistical attacks on the list by not only using decoy values, but flipping those bits randomly (when you flip other bits in the list).

msporny:
PR #155 has been raised to address this issue. This issue will be closed once PR #155 has been merged.

msporny:
PR #155 has been merged, closing.

npdoty:
@dlongley and others have argued that there are privacy harms and no privacy benefits to decoy entries, and the group has suggested that it is unaware of any threat or use case where a decoy value could provide a privacy benefit.

To be clear, I don't think of this as an "obvious" solution at all and I'm not as familiar with these deployments. But I suspect there are threats where decoy values could provide some protection, especially when status value changes are rare, where the group of people with a particular status is small or where other information is known about people whose status may be on the list.

For example, an issuer very rarely suspends licenses for a particular behavior. A case documenting that behavior is widely published in the press and on the same day, the issuer updates the status list to indicate a suspended status. The press then report apparent confirmation that the suspect's license was suspended, and potentially other information about them (when it was last suspended or un-suspended), even if that information was intended to be kept confidential. If the status information was shared in cases of selective disclosure (the licensee had proved their license status in order to access sensitive content online), then the licensee's identity has also been disclosed, and the site learns the identity of the visitor who accessed that particular content.

@KDean-Dolphin raised a separate use case about business intelligence, where the size of the status list might reveal the group behavior, like how many licenses are being issued during a particular timeframe, that the issuer might wish to conceal.

npdoty:
Yes, I think noting it as an open question on how to do properly (or whether), with an at-risk marker, would make sense for CR.

msporny:
PR #171 has been raised to note that the decoy guidance will be refined during the Candidate Recommendation process and that the group may, or may not, suggest that decoy values are good/bad/a mixed bag.

[...]

PR #155, opened 2024-03-24
Add privacy considerations section about decoy values.

msporny:
This PR is an attempt to address issue #150 by providing guidance around decoy values.

/cc @zoracon

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/155.html" title="Last updated on Apr 6, 2024, 5:57 PM UTC (331d95f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/155/9ee8c7e...331d95f.html" title="Last updated on Apr 6, 2024, 5:57 PM UTC (331d95f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1168,46 +1168,46 @@ from <a>issuers</a>.
       <h3>Decoy Values</h3>
 
       <p>
-The use of decoy values in status lists by <a>issuers</a> has been explored as
-a mechanism to increase the privacy of <a>subjects</a> by further hiding the
-size of the group associated with a status list as well as when values in the
+[=Issuer=] use of decoy values in status lists has been explored as
+a mechanism to increase the privacy of [=subjects=] by further hiding both
+the size of the group associated with a status list and when values in the
 list change. While algorithms for employing decoy values are out of scope for
 this specification, implementers are advised to heed the guidance in this
 section when considering the use of decoy values.
       </p>
       <p>
 Decoys might help privacy in some cases, but harm privacy in others. For
-example, if status list entry indexes are allocated in a purely random fashion,
+example, if status list entry indexes are allocated in a random fashion,
 adding decoys harms privacy because it reduces the group privacy size by the
-number of decoys added to the group. A purely random allocation of indexes
-ensures that decoys are not necessary unless a bulk operation, such as setting
-every credential issued in a list to the same status (such as "revoked"),
-would reveal the true size of the group.
+number of decoys added to the group. A random allocation of indexes inherently
+hides the true group size, ensuring that decoys are not necessary unless a bulk
+operation, such as setting every credential in a list to the same status
+(such as "revoked"), could reveal the true size of the group to an adversary
+that also knew that the bulk update was to be applied to all non-decoy indexes.```
       </p>
       <p>
-If decoys are used, the proper number of decoys to use is a function that
-includes at least the desired group privacy size, the randomness of the
-distribution of entries in the set, and ensuring that entities watching the list
-cannot determine which entries are real and which ones are decoys as they change
-throughout time. These variables often change based on the use case and
-implementers are encouraged to carefully evaluate whether using decoys would
-help or harm privacy for their particular scenario.
+If decoys are used, the proper number of decoys to use is a function of
+at least the desired group privacy size, the randomness of the
+distribution of entries in the set, and whether entities watching the list
+can determine which entries are real and which are decoys as they change
+through time. These variables often change based on the use case and
+implementers are encouraged to carefully evaluate whether decoys would
+help or harm privacy for their particular deployment scenario.
       </p>
       <p>
-When employing decoy values, it is important to ensure that the decoys
+When using decoy values, it is important to ensure that the decoys
 behave like real entries in the group. For example, if a decoy is used for
 revocation, it would be strange to flip the decoy value to "unrevoked" when that
-is not how the rest of the revocation values associated with real entries
-behave. Similarly, changing the decoy values at times and numbers that do not
-align with the way real entries behave can enable statistical analysis to
-separate the real entries from the decoy entries.
+is not how the revocation values associated with real entries
+behave. Similarly, changing the decoy values at times and to values that do not
+mimic the behavior of real entries can allow the real entries to be separated
+from the decoy entries through statistical analysis.
       </p>
       <p>
-In general, the use of decoys is discouraged as, more often than not,
-random allocation of status list entry indexes provides adequate protection
-for most use cases. If bulk operations can be performed on a status list
-that would reveal the group size, only then might an implementer consider the
-use of decoys.
+The use of decoys is discouraged for most use cases, as random allocation
+of status list entry indexes provides adequate protection.
+Decoy are only to be considered in deployments where bulk
+operations that would reveal the group size can be performed on a status list.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -1168,18 +1168,26 @@ from <a>issuers</a>.
       <h3>Decoy Values</h3>
 
       <p>
-The use of decoy values in status lists by <a>issuers</a> can increase the
-privacy of <a>subjects</a> by further hiding the size of the group associated
-with a status list as well as when values in the list change. When employing
-decoy values, it is important to ensure that the decoys behave like real
-entries in the group. For example, if a decoy is used for revocation, it would
-be strange to flip the decoy value to "unrevoked" when that is not how the
-rest of the revocation values associated with real entries behave. Similarly,
-changing the decoy values at times and numbers that do not align with the way
-real entries behave can enable statistical analysis to separate the real
-entries from the decoy entries. While algorithms for employing decoy values
-are out of scope for this specification, the usage of decoy values is an
-important part of an <a>issuer's</a> privacy protections for their
+The use of decoy values in status lists by <a>issuers</a> can, when used
+properly, increase the privacy of <a>subjects</a> by further hiding the size of
+the group associated with a status list as well as when values in the list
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1149,6 +1149,26 @@ <h3>Content Distribution Networks</h3>
       </p>
     </section>
 
+    <section class="informative">
+      <h3>Decoy Values</h3>
+
+      <p>
+The use of decoy values in status lists by <a>issuers</a> can increase the
```
dlongley, line 1156: Decoy values, which are known to the issuer, also decrease the size of the group -- which is relied upon for privacy. It might be the case that random assignment of indexes in a list provides sufficient protection against whatever decoys might do, without reducing the size of the group. I'm not convinced that decoys add more value in many cases (such as for revocation status lists for which there are few revocations) than they would take away.

It might be better to say nothing about them here unless there is further analysis / text explaining the threats that they mitigate that pseudo-random assignment of indexes does not.

index.html
```diff
@@ -1149,6 +1149,26 @@ <h3>Content Distribution Networks</h3>
...
+privacy of <a>holders</a> by further hiding the size of the group associated
```
David-Chadwick, line 1157:
```suggestion
privacy of <a>subjects</a> by further hiding the size of the group associated
```

index.html
```diff
@@ -1149,6 +1149,26 @@ <h3>Content Distribution Networks</h3>
...
+decoy values, it is important to ensure that the decoys behave like real
+entries in the group. For example, if a decoy is used for revocation, it would
+be strange to flip the decoy value to "unrevoked" when that is not how the
+rest of the revocation values associated with real entries behave. Similarly,
+changing the decoy values at times and numbers that do not align with the way
+real entries behave can enable statistical analysis to separate the real
+entries from the decoy entries. While algorithms for employing decoy values
+are out of scope for this specification, the usage of decoy values is an
+important part of an <a>issuer's</a> privacy protections for their
+<a>holders</a>.
```
David-Chadwick, line 1168:
```suggestion
<a>subjects</a>.
```

David-Chadwick (changes requested):

decentralgabe (approved):

index.html
msporny: Yes, I see your point.

I tried some updates in aada152a9bf66b21f529311ebd22d10e1a5d02ca and efa14d9c04f7d8216b25dcb42f1d66829071d2fa to address your concerns.

We need to say /something/ about decoys because they are being proposed as an "obvious solution" in some digital credentials initiatives even though there are known privacy harms when using them.

Any thoughts on the updated text?

index.html
```diff
@@ -1164,6 +1164,53 @@ <h3>Content Distribution Networks</h3>
...
+section when considering the use of decoy values.
+      </p>
+      <p>
+Decoys might help privacy in some cases, but harm privacy in others. For
+example, if status list entry indexes are allocated in a purely random fashion,
+adding decoys harms privacy because it reduces the group privacy size by the
+number of decoys added to the group. A purely random allocation of indexes
+ensures that decoys are not necessary unless a bulk operation, such as setting
+every credential issued in a list to the same status (such as "revoked"),
+would reveal the true size of the group.
```
dlongley, line 1185: Tweaked some language here, but I'm still having trouble finding the value of decoys in the "bulk update" case (again, presuming random allocation of indexes is already used).

I don't think I fully understand the threat model. I understand that it could be problematic for an adversary to learn the "true group size" when it is significantly smaller than the total size of the list. Learning this would allow an adversary to make more accurate guesses about the use of a particular VC if they also:

1. knew of at least one VC associated with the list, and
2. could observe retrieval of the list by other parties.

Is there a problem if an adversary learns the "true group size" is close to or the same size as the total size of the list (for a list with over ~100K indexes)? Is it a goal for an adversary to simply never learn how many total VCs have been issued -- and, if so, is this still about subject privacy or is it about issuer privacy? What if they learn N out of M have been issued? Is that ok -- where is the line? Is "the total number" special for some reason?

Further analyzing this "bulk update" case, there is of course a presumption that the adversary does not already know the "true group size" (since this is the threat). And my understanding is that a "bulk update" means that an adversary could observe N-many bits flip at once (for N > 1).

Now supposing that an adversary also knows that there are no decoys, how would simply observing any number of N bits change (where N is less than the total list size) tell them the "true group size"? How are they to know whether a "bulk update" changed all of the statuses of the "true group size" and not just some subset of it?

Is there an additional presumption that they also know that, for example, all VCs in that particular list (or perhaps, all VCs for that issuer across all lists) are about to be revoked -- and they are just waiting to see how many bits flip? Why would an adversary presume there is a single list or that they have knowledge of (and an address for) all current lists such that they could observe them all?

For this threat to be realized, it seems there may be a considerable amount of information that the adversary needs to have -- all whilst not being able to make a good guess at the "true group size" using that very same information.

```suggestion
number of decoys added to the group. A random allocation of indexes already
hides the true group size, ensuring that decoys are not necessary unless a bulk
operation, such as setting every credential issued in a list to the same status
(such as "revoked"), would reveal the true size of the group to an adversary
that also knew that the bulk update was to be applied to all non-decoy indexes.
```

index.html
```diff
@@ -1164,6 +1164,53 @@ <h3>Content Distribution Networks</h3>
...
+revocation, it would be strange to flip the decoy value to "unrevoked" when that
+is not how the rest of the revocation values associated with real entries
+behave. Similarly, changing the decoy values at times and numbers that do not
+align with the way real entries behave can enable statistical analysis to
+separate the real entries from the decoy entries.
+      </p>
+      <p>
+In general, the use of decoys is discouraged as, more often than not,
+random allocation of status list entry indexes provides adequate protection
+for most use cases. If bulk operations can be performed on a status list
```
dlongley, line 1208: This first sentence is the main thing I agree with in this PR. Don't use decoys, trying to do it seems like you'll probably just end up harming privacy for little to no gain. The second sentence I can't really say I agree with at this point in time. I remain unconvinced by the text that decoys are a good idea in any circumstance when randomness is already employed. I guess if you can't do your index allocation randomly enough, decoys could be of some help, so I wouldn't block the PR on that basis -- but if there's consensus to remove mention of decoys entirely I'd prefer that.

It would be ideal for a "decoy advocate" to offer better text in support of their position; I understand we're trying to approximate here for something that is meant to be obvious but maybe it isn't so obvious in this circumstance. I understand the basic strategy and concept behind using decoys to hide real information in a system, but when this comes into contact with the group privacy problem and other potential solutions here -- I'm not sure it survives analysis. I'm happy to be wrong and would prefer to see the text making it clear to implementers when/why to use them, because that's what this section is for -- and I think a number of them would get a little lost here.

index.html
```diff
@@ -1164,6 +1164,53 @@ <h3>Content Distribution Networks</h3>
...
+The use of decoy values in status lists by <a>issuers</a> has been explored as
+a mechanism to increase the privacy of <a>subjects</a> by further hiding the
+size of the group associated with a status list as well as when values in the
+list change. While algorithms for employing decoy values are out of scope for
+this specification, implementers are advised to heed the guidance in this
+section when considering the use of decoy values.
+      </p>
+      <p>
+Decoys might help privacy in some cases, but harm privacy in others. For
+example, if status list entry indexes are allocated in a purely random fashion,
```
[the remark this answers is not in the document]
msporny: I don't think we'd want to suggest that people mix the two approaches... the best practice is "random". Doing anything other than that is asking for trouble.

msporny:
@dlongley @TallTed I have reworded this PR to discourage the use of decoy values (as randomization seems to be the simplest approach that maximizes privacy). We'll need an advocate for decoy values to provide a use case for why they might be useful in certain use cases. Until then, the specification will encourage the use of randomization and discourage the use of decoy values.

msporny:

index.html
```diff
@@ -1168,46 +1168,25 @@ <h3>Content Distribution Networks</h3>
...
-a mechanism to increase the privacy of [=subjects=] by further hiding both
-the size of the group associated with a status list and when values in the
-list change. While algorithms for employing decoy values are out of scope for
-this specification, implementers are advised to heed the guidance in this
-section when considering the use of decoy values.
+[=Issuer=] use of decoy values in status lists has been explored as a mechanism
+to increase the privacy of [=subjects=]. While algorithms for employing decoy
+values are out of scope for this specification, implementers are advised that
+the use of decoy values do not provide privacy gains and can harm privacy in
+most cases.
```
TallTed, line 1175: Anything merged to main without review is liable to have lingering issues. Now I gotta make another PR...

index.html
```diff
@@ -1168,46 +1168,25 @@ <h3>Content Distribution Networks</h3>
...
-example, if status list entry indexes are allocated in a random fashion,
-adding decoys harms privacy because it reduces the group privacy size by the
-number of decoys added to the group. A random allocation of indexes inherently
-hides the true group size, ensuring that decoys are not necessary unless a bulk
-operation, such as setting every credential in a list to the same status
-(such as "revoked"), could reveal the true size of the group to an adversary
-that also knew that the bulk update was to be applied to all non-decoy indexes.```
+When status list entry indexes are allocated in a random fashion, which is the
+suggested mode of operation for this specification, adding decoys harms privacy
+because it reduces the group privacy size by the number of decoys added to the
```
TallTed, line 1180: What is a "group privacy size"? I could find no suitable definition on the web, making it seem likely to have been invented here, where there also seems to be no definition. Once there is a definition I can refer to, I expect this paragraph to need some rephrasing.

TallTed: pick up in #166

index.html
TallTed: #166

Issue #151, opened 2024-03-13
Remove open ended status messages because it is a layer violation


jandrieu:
In discussing issue #146, @msporny clarified that the current bitstring status list supports arbitrary messages in status lists.

Unfortunately, this turns status lists into yet another attestation architecture, which is a layering violation. Status lists don't have the privacy and security features we have for VCs, which is where, in the VC architecture, issuers make formal attestations about subjects.

Bitstring status lists should be used to enable decentralized verification of pre-arranged possible statuses, not of arbitrary, potentially changing attributes of the subject.

msporny:
@jandrieu Given that the specification has already marked the status messages feature as "at risk", and is seeking implementer feedback during CR, do you see this issue as potentially blocking our transition to CR next month?

brentzundel:
Isn't status returned as a VC? That would mean Status lists are using the VC attestation architecture which has precisely the privacy and security features we have for VCs.

The trouble with our WG trying to pre-arrange every possible status is that we can't possibly get them all. Why shouldn't a status VC have extension points just like a regular VC?

Perhaps it needs to be clarified that the subject of a status VC is always another credential, therefore any messages in a status VC are only ever messages about the status of the VC, not about the subject of the original VC.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-03-27-vcwg#section2-1) on 2024-03-27

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.1. Remove open ended status messages because it is a layer violation (issue vc-bitstring-status-list#151)

_See github issue [vc-bitstring-status-list#151](https://github.com/w3c/vc-bitstring-status-list/issues/151)._




**Manu Sporny:** PING have raised several privacy issues that we should address.  
… several issues around I18N that need to be reviewed.  
… strings may be human readable.  

> *Manu Sporny:* [https://github.com/w3c/vc-data-model/pulls](https://github.com/w3c/vc-data-model/pulls).

**Manu Sporny:** will be a massive PR to address Yaskin's comments.  
… mostly editorial apart from a few normative statements (but no new implementation requirements).  

</details>

jandrieu:
> Isn't status returned as a VC? That would mean Status lists are using the VC attestation architecture which has precisely the privacy and security features we have for VCs.

[...]

PR #164, opened 2024-04-06
Ensure that status messages are committed to at issuance time.

msporny:
This PR is an attempt to address issue #151 by requiring that status messages are committed to by issuers at issuance time. The PR does this by:

* Moving the `statusSize`, `statusMessage`, and `statusReference` fields to the `BitstringStatusListEntry` object.
* Explaining why the commitment is done at issuance time on the original VC itself.
 
NOTE TO REVIEWERS: Most of the changes in this PR are just moving text around in the spec. The only normative change is the first bullet item above. Everything else should be editorial. It looks like way more changed than it did.

I believe, if this PR is accepted, that it should address @jandrieu and PINGs concerns regarding the issuer changing the types of status messages that are possible without the holder or subjects knowledge.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/164.html" title="Last updated on Apr 16, 2024, 8:02 PM UTC (93a020b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/164/6626331...93a020b.html" title="Last updated on Apr 16, 2024, 8:02 PM UTC (93a020b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -624,10 +624,18 @@ a credential may involve some understanding of the business case involved.
 
 
         <p>
-Status list entries can be used to associate a single status type,
-such as `revocation` or `suspension`, with a [=verifiable credential=] by using
-the `statusPurpose` property. The example below demonstrates the association
-of simple status list entries:
+Status list entries can be used to express the purpose of a
+status associated with a [=verifiable credential=] by using the
+`statusPurpose` property.
+        </p>
+        <p>
+The use of `revocation` or `suspension` as the status purpose
+includes the semantics of the status, with `revocation`
+indicating that a status bit expresses whether a
+[=verifiable credential=] has been revoked and `suspension`
+indicating that a status bit expresses whether a
+[=verifiable credential=] has been suspended. The example below
+demonstrates the use of these status purposes:
         </p>
 
         <pre class="example nohighlight"
--- a/index.html
+++ b/index.html
@@ -465,6 +465,17 @@ expression of the data model in this section MUST be expressed in a
 conforming [=verifiable credential=] as defined in [[VC-DATA-MODEL-2.0]].
         </p>
 
+        <p class="issue atrisk"
+           title="Bitstring entry sizes greater than 1 are at risk.">
+The Working Group is currently seeking implementer feedback regarding the
+utility of bitstring entries that have sizes greater than one. Supporting
+such entries adds complexity to the solution, and it's not clear whether there
+is enough of an implementation community to support the feature. The WG is
+considering three options: (1) require conforming implementations to support
+the feature; (2) allow implementations to optionally support the feature; or
+(3) remove the feature. At present, the specification implements option (2).
+        </p>
+
         <table class="simple">
           <thead>
             <tr>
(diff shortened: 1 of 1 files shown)


David-Chadwick (approved):

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
+status codes.
+              </td>
+            </tr>
+            <tr>
+              <td id="statusReference">statusReference</td>
+              <td>
+The `statusReference` property provides a point for implementers to include a
+[[URL]] to material related to the status. An implementer MAY include the
+`statusReference` property, and if they do, the value MUST be a [[URL]] or an
+array of URLs. Implementers using a `statusPurpose` of `status` are strongly
```
dlongley, line 614:
```suggestion
array of URLs. Implementers using a `statusPurpose` of `message` are strongly
```

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
+If present, elements in the `statusMessage` array MUST contain at
+minimum two properties:
+                <ul>
+                  <li id="status">
+`status`, a string representing the hexadecimal value of the status prefixed
+with `0x`
+                  </li>
+                  <li id="message">
+`message`, a string used by software developers to assist with debugging
+which SHOULD NOT be displayed to end users.
```
dlongley, line 596: How can these messages be committed to for users if they are not to be displayed or understood by them?

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
+Implementers MAY use the string value of `undefined` in the value to indicate
+that a corresponding status is not defined for the associated status value, but
+that it may be defined in the future. Rules for how to handle various status
+messages are outside the scope of normative requirements in this document, but
+it is assumed that implementers will document rules for processing various
+status codes.
+              </td>
+            </tr>
+            <tr>
+              <td id="statusReference">statusReference</td>
```
dlongley, line 609: Does `statusReference` pose a privacy problem in the same way that uncommitted messages might? Perhaps a hash is also required (e.g., `statusReference: {id: "https://example.com/foo", digestMultibase: "<hash>"}`?

If `statusReference` were committed to via a hash, the `statusMessage` array could potentially be omitted entirely. This might be more attractive than always including all possible messages, now that they must be present in the VC itself. This may also help address the i18n + commitment issue (that I think might exist now as mentioned above) because a more rich response can be returned in the status reference document.

dlongley (changes requested):

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
+
+        <p>
+Status list entries can be used to associate many status types with a
+[=verifiable credential=] by using the `message` status purpose. The set of
+messages associated with a particular entry is committed to by the [=issuer=]
+by using the `statusSize`, `statusMessage`, and optional `statusReference`
+properties. This commitment is done at the time of issuance to ensure that
+the [=holder=] knows what sort of information might be associated with a
+particular [=verifiable credential=] that is in their possession, and would
+then be discoverable by a [=verifier=] that receives that credential.
```
TallTed, line 673:
```suggestion
Status list entries can be used to associate multiple status types with a
[=verifiable credential=] through the single `message` status purpose. The
[=issuer=] commits to the set of messages that may be associated with a
particular entry (i.e., with a particular [=verifiable credential=]) through
the `statusSize`, `statusMessage`, and optional `statusReference` properties,
at the time of [=verifiable credential=] issuance. This is to ensure that
the [=holder=] knows what sort of information might be associated with a
particular [=verifiable credential=] they keep in their possession, that
could then be discoverable by a [=verifier=] that later receives that
credential.
```

index.html
```diff
@@ -594,20 +716,29 @@ <h3>BitstringStatusListCredential</h3>
...
-the feature; (2) allow implementations to optionally support the feature; or
-(3) remove the feature. At present, the specification implements option (2).
+        <p>
+The status list is expressed inside a [=verifiable credential=] in order to
+enable a [=holder=] to provide it to a [=verifier=] directly. This mechanism,
+sometimes called "certificate stapling", increases privacy for the [=holder=] by
+ensuring that the [=verifier=] does not need to contact the [=issuer=] to
+retrieve the status list. Still, a [=verifier=] might choose to ignore the
+[=holder=]-provided status list, even when its authenticity is verifiable,
+if it desires a more recent version of a status list, for instance.
```
TallTed, line 729:
```suggestion
The status list is expressed inside a [=verifiable credential=] in order to
enable a [=holder=] to provide it directly to a [=verifier=]. This mechanism,
sometimes called "certificate stapling", increases privacy for the [=holder=] by
ensuring that the [=verifier=] does not need to contact the [=issuer=] to
retrieve the status list. Still, a [=verifier=] might choose to ignore the
[=holder=]-provided status list, even when its authenticity is verifiable,
if it desires a more recent version of a status list, for instance.
```

index.html
```diff
@@ -594,20 +716,29 @@ <h3>BitstringStatusListCredential</h3>
...
 
+        <p>
+[=Issuers=] and [=verifiers=] are advised that the [=issuer=] of a
+[=verifiable credential=] and the [=issuer=] of an associated
+`BitstringStatusListCredential` might not be the same. There are technical,
+legal, institutional, and political reasons that might make it appropriate
+to separate the authority over the original credential from the authority to
+revoke such a credential. Therefore, the `issuer` value of a <a>verifiable
+credential</a> containing a `BitstringStatusListEntry` MAY be different from
+the `issuer` value of a `BitstringStatusListCredential`.
```
TallTed, line 740:
```suggestion
[=Issuers=] and [=verifiers=] are advised that the [=issuer=] of a
[=verifiable credential=] and the [=issuer=] of an associated
`BitstringStatusListCredential` might not be the same. There are technical,
legal, institutional, and political reasons that might make it appropriate
to separate the authority over the original credential from the authority to
revoke, or otherwise change the status of, such a credential. Therefore, the
`issuer` value of a [=verifiable credential=] containing a
`BitstringStatusListEntry` MAY be different from the `issuer` value of a
`BitstringStatusListCredential`.
```

TallTed (changes requested):

brentzundel (approved):
approved with Ted's suggested changes

mkhraisha (approved):

decentralgabe (approved):

brianorwhatever (approved):

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
+              <td>
+The `statusMessage` property MUST be an array. If present,
+the length of the array must equal the number of possible status states
+indicated by `statusSize`. `statusMessages` MAY be present if
+`statusSize` is `1`. `statusMessages` MUST be present if
+`statusSize` is greater than `1`.  If not present, the message value
+associated with the bit value of `0` is "unset" and the bit
+value of `1` is "set".
+If present, elements in the `statusMessage` array MUST contain at
+minimum two properties:
```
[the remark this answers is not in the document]
dlongley: > (i.e., the `statusMessage` array MUST have as many elements as there are bits in `statusSize`).

Hmm, I would think it needs to have as many elements as there are possible values instead, right? For example, if there are 2 bits, that's 4 possible values -- so 4 elements are required, not 2.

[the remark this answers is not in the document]
TallTed: Ah, yes, good catch! Fixing...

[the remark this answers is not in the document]
TallTed: @dlongley — Not a trivial reword. See if this makes sense —
```suggestion
If present, the `statusMessage` property MUST be an array, the length
of which MUST equal the number of possible status states indicated by
`statusSize` (e.g., `statusMessage` array MUST have 2 elements if
`statusSize` has 1 bit, 4 elements if `statusSize` has 2 bits, 8
elements if `statusSize` has 3 bits, etc.). `statusMessage` MAY be
present if `statusSize` is `1`, and MUST be present if `statusSize` is
greater than `1`.  If the `statusMessage` array is not present, the
message values associated with the `status` bit values of `1` and `0`
are "set" and "unset", respectively. If the `statusMessage` array is
present, each element MUST contain the two properties described below,
and MAY contain additional properties.
```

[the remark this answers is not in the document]
TallTed: > "possible status messages"? Since this is for the "message" status purpose type?

The above, applied below. 

I don't have any solution bound thoughts on the other points raised [here](https://github.com/w3c/vc-bitstring-status-list/pull/164#discussion_r1558312457) (which maybe should be raised in a new, independent issue?)
```suggestion
If present, the `statusMessage` property MUST be an array, the length
of which MUST equal the number of possible status messages indicated
by `statusSize` (e.g., `statusMessage` array MUST have 2 elements if
`statusSize` has 1 bit, 4 elements if `statusSize` has 2 bits, 8
elements if `statusSize` has 3 bits, etc.). `statusMessage` MAY be
present if `statusSize` is `1`, and MUST be present if `statusSize` is
greater than `1`.  If the `statusMessage` array is not present, the
message values associated with the `status` bit values of `1` and `0`
are "set" and "unset", respectively. If the `statusMessage` array is
present, each element MUST contain the two properties described below,
and MAY contain additional properties.
```

index.html
mprorock: could we just treat this identical to an external reference as in the core data model?

dlongley: I think that's the basic idea, it was just being given a more specific relationship name since it's well known how it relates.

index.html
```diff
@@ -552,39 +563,150 @@ <h3>BitstringStatusListEntry</h3>
...
           </tbody>
         </table>
 
-        <pre class="example nohighlight" title="Example StatusListCredential">
+
+        <p>
+Status list entries can be used to associate a single status type,
+such as `revocation` or `suspension`, with a [=verifiable credential=] by using
+the `statusPurpose` property. The example below demonstrates the association
+of simple status list entries:
```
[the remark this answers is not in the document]
dlongley: "a status type" => "a status purpose"?

index.html
dlongley: So "message" is just another "status purpose", it doesn't enable "multiple status types" (I don't think). It is just the case that just about any "message" about the status of a credential can be expressed. The purpose of a "message" status list expands to: "to express a specific message (one from a list of precommitted messages) describing the status of a credential". We should probably just avoid saying "status type" and stick to "status purpose".

So we have:

1. A status purpose of "revocation", which means that a status list is used to express the revocation status of a particular credential.
2. A status purpose of "suspension", which means that a status list is used to express the suspension status of a particular credential.
3. A status purpose of "message", which means that a status list is used to express a specific message, from a predefined list of messages, describing the status of a particular credential.

dlongley: Maybe some text like this can be incorporated into the suggestions above:

```
Status list entries can be used to express the purpose of a
status associated with a [=verifiable credential=] by using the
`statusPurpose` property.

The use of `revocation` or `suspension` as the status purpose
includes the semantics of the status, with `revocation`
indicating that a status bit expresses whether a
[=verifiable credential=] has been revoked and `suspension`
indicating that a status bit expresses whether a
[=verifiable credential=] has been suspended:

/* revocation and suspension examples */

The use of `message` as the status purpose enables an issuer to
define an arbitrary number of custom, descriptive messages about
the status of the [=verifiable credential=]. The [=issuer=] commits...
...

File: index.html (hunk 8 above)
Comment (dlongley) [inline index.html:630]: My attempt to incorporate the suggestions here together:

```suggestion
        <p>
Status list entries can be used to express the purpose of a
status associated with a [=verifiable credential=] by using the
`statusPurpose` property.
        </p>
        <p>
The use of `revocation` or `suspension` as the status purpose
includes the semantics of the status, with `revocation`
indicating that a status bit expresses whether a
[=verifiable credential=] has been revoked and `suspension`
indicating that a status bit expresses whether a
[=verifiable credential=] has been suspended:
```

index.html
dlongley, line 673: My attempt to incorporate everything above:

```suggestion
The use of `message` as the status purpose enables an issuer to
define an arbitrary number of custom, descriptive messages about
the status of the [=verifiable credential=]. The [=issuer=] commits to
the set of messages that may be associated with a particular entry
(i.e., with a particular [=verifiable credential=]) through the
`statusSize`, `statusMessage`, and optional `statusReference` properties,
at the time of [=verifiable credential=] issuance. This is to ensure that
the [=holder=] knows what sort of information might be associated with a
particular [=verifiable credential=] they keep in their possession, that
could then be discoverable by a [=verifier=] that later receives that
credential.
```

dlongley (changes requested):

index.html
dlongley: I added a suggestion below that attempts to combine Ted's suggestion with my own.

index.html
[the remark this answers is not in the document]
dlongley: My comment below includes a suggestion that incorporates Ted's suggestion with some edits.

index.html
msporny: There is a presumption here that the user software will convert the value into something internationalized for the individual based on its programmatic value.

I do think that's excessively hand-wavy, given that the issuer could just do that work here, but given that the Traceability group wants it to work this way, and that there are no other strong opinions on how it should work, the text that's in the spec is the best we have at present.

index.html
msporny: I thought the `statusReference` was human-readable documentation, and so committing to it using a hash would prevent it being updated in ways that would harm readability?

Again, the design here is strange in that we have a unique value for each message (`statusMessage.status`), but then we have another "developer value" (`statusMessage.message`), that could just as easily live in `statusReference` (if it were a machine-readable file).

It's up to Traceability folks if they want to change the design to something simpler. I'm not going to let that hold up this PR.

msporny:
Normative, multiple reviews, changes requested and made, no objections (but non-blocking design concerns raised), merging.

Issue #183, opened 2024-11-15
New status purpose proposal


PatStLouis:
Greetings, while implementing `BitstringStatusList` in our use case (supply-chain), we uncovered a situation where an issued credential might have a new version available for pickup which doesn't warrant a revocation of the previous version, but it would still be worthwhile to have a status to signal that an updated version is available.

Therefore I would like to suggest adding a status purpose ~`supersession`~ `refresh`, which can have a value of `0`(meaning this is the latest version of the credential available) or `1` (MAY be refreshed).

The process for a holder and/or verifier to get the latest version is out of scope, however I would intend this to be used in parallel with a ~`SupersessionRefresh`~ defined `refreshService` type, which contains an endpoint to query and get an updated version of the credential.

Once the software have retrieved the latest version of the credential, they would archive the current version and replace it with the latest version in their software, or create a [Symbolic link](https://en.wikipedia.org/wiki/Symbolic_link) towards it.

edit: renamed `supersession` to `refresh` as proposed.

msporny:
+1 for this feature. 

Can we name the status purpose "refresh" instead of "supersession"? The former might be easier to understand.

PatStLouis:
refresh would make sense, especially if its paired with a refresh service (optional, could also be out of bound). I had `update` at first then changed to `supersession` to align with the revocation/suspension terminologies.

Whichever term has consensus around which is representative of the features.

msporny:
PR #185 has been raised to address this issue. This issue will be closed once PR #185 has been merged.

msporny:
PR #185 has been merged, closing.

PR #185, opened 2024-11-24
Add `refresh` status purpose.

msporny:
This PR is an attempt to address issue #183 by adding a `refresh` status purpose.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/185.html" title="Last updated on Nov 30, 2024, 6:09 PM UTC (4908579)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/185/8af361e...4908579.html" title="Last updated on Nov 30, 2024, 6:09 PM UTC (4908579)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -876,6 +876,14 @@ values MUST be used for their intended purpose:
                   </thead>
                   <tbody>
                     <tr>
+                      <td>`refresh`</td>
+                      <td>
+Used to signal that an updated [=verifiable credential=] is available via the
+credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">Refresh service</a>
+feature. This status is not reversible.
+                      </td>
+                    </tr>
+                    <tr>
                       <td>`revocation`</td>
                       <td>
 Used to cancel the validity of a [=verifiable credential=]. This status is
--- a/index.html
+++ b/index.html
@@ -880,7 +880,7 @@ values MUST be used for their intended purpose:
                       <td>
 Used to signal that an updated [=verifiable credential=] is available via the
 credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">refresh service</a>
-feature. This status is not reversible.
+feature. This status does not invalidate the [=verifiable credential=] and is not reversible.
                       </td>
                     </tr>
                     <tr>
--- a/index.html
+++ b/index.html
@@ -879,7 +879,7 @@ values MUST be used for their intended purpose:
                       <td>`refresh`</td>
                       <td>
 Used to signal that an updated [=verifiable credential=] is available via the
-credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">Refresh service</a>
+credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">refresh service</a>
 feature. This status is not reversible.
                       </td>
                     </tr>


index.html
```diff
@@ -875,6 +875,14 @@ <h3>BitstringStatusListCredential</h3>
                     </tr>
                   </thead>
                   <tbody>
+                    <tr>
+                      <td>`refresh`</td>
+                      <td>
+Used to signal that an updated [=verifiable credential=] is available via the
+credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">Refresh service</a>
```
TallTed, line 882: lowercase?
```suggestion
credential's <a data-cite="VC-DATA-MODEL-2.0#refreshing">refresh service</a>
```

iherman (approved):
The new property should be added, I presume, to the bitstring vocabulary document, and to the diagram include therein.

It is o.k. if this is handled in a separate Issue+PR, but it ought to be done. Please raise an issue if you prefer to handle it separately.

Similarly, shouldn't the relevant context file be changed?

(I am happy to do the changes on the vocabulary... eventually. I am not at my computer right now, it would be difficult to do it on my tablet...)

msporny:
> The new property should be added, I presume, to the bitstring vocabulary document, and to the diagram include therein.

Nope, it's not a property, it's a string. :)

PatStLouis:
I think we can have the purpose added and explore discussions around if it needs an additional defined property separately. I think if a new term needs to be defined it would be more towards an associated `refreshService`, which can live in a separate context.

@msporny Would it be reasonable to add to this text that this refresh status doesn't have an effect on the validity of the current credential, meaning it doesn't become revoked because there's an update available, simply signals that there's an update. Revocation should still be managed by a separate status entry. If there's a "breaking update", the credential would get updated + revoked. If its a "soft" (meaning a minor change) or "additive"(adding a data element) update, the credential gets updated, but the consumer of the credential can still use the non updated version without concern (they don't need to get the updated version). If its also revoked then they would need to get the updated version if they want to continue using the data for their services.

I think just a sentence that the refresh status doesn't invalidate the current credential.

PatStLouis (approved):

mkhraisha (approved):

PatStLouis:
@TallTed the purpose is that you would resolve the index of the refresh status, 0 means no update available and 1 means an update is available you MAY hit the refresh endpoint, so it does signal that there's an update available.

iherman:
> > The new property should be added, I presume, to the bitstring vocabulary document, and to the diagram include therein.
> 
> Nope, it's not a term, it's an arbitrary string. :)

Oops, sorry.

(For some reasons preview does not work, and I have only my ipad with me where I am this week, so it was difficult to see the exact context of the changes. My approval holds :-)

index.html
```diff
@@ -875,6 +875,14 @@ <h3>BitstringStatusListCredential</h3>
...
+feature. This status is not reversible.
```
brianorwhatever, line 883:
```suggestion
feature. This status does not invalidate the [=verifiable credential=] and is not reversible.
```

brianorwhatever (approved):
Something like this to address the comments?

longpd (approved):
I approve this PR with the amended language it contains.

Jsyro (approved):

msporny:

Issue #48, opened 2023-01-31
Document design characteristics why "StatusList2021Credential" is a VC


Sakurann:
... because if it is a status of a VC, entities are already expected to support VCs.

One advantage of it being a VC is that StatusList2021Credential can be downloaded by the Holder and sent to the Verifier when Holder is offline. Though if I am an attacker, I will download a version before my VC gets revoked, and keep sending it... so security considerations that a verifier needs to be careful when accepting a StatusList2021Credential offline should be included. Or include a "statusListCredential" URL inside statusList2021Credential so that there is a circular logic?

dlongley:
The security considerations should indicate at least two things to help verifiers mitigate problems here:

1. Advise verifiers not to accept a status list VC that is outside of its validity period.
2. If the status list VC validity period is "too large" (specify something for this and the rationale), advise verifiers not to accept a status list VC that is "too large"-many minutes after the beginning of the validity period if they did not retrieve it directly from the issuer themselves.

Additionally, advice should be given to issuers to have validity periods no larger than "too large" and to expect that verifiers will follow the above advice.

msporny:
PR #162 has been raised to partially address this issue by documenting why the status list is expressed in a VC.

msporny:
PR #162 has been merged, the tasks remaining to resolve this issue are:

* Provide guidance for acceptable VC validity periods, and what to do if the period is too large.

msporny:
PR #190 has been raised to address this issue. This issue will be closed once PR #190 has been merged.

msporny:
PR #190 has been merged, closing.

PR #190, opened 2024-11-30
Provide guidance on choosing validity periods for status lists.

msporny:
This PR is an attempt to address issue #48 by providing guidance on how to choose validity periods for status lists.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Dec 2, 2024, 7:00 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL]([object Object])

```
Protocol error (Runtime.callFunctionOn): Promise was collected
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/vc-bitstring-status-list%23190.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1682,12 +1682,13 @@ be 0, and the last index will be 131,071.
       <h3>Validity Periods</h3>
 
       <p>
-The <a data-cite="?VC-DATA-MODEL-2.0#validity-period">validity period</a> of a
-status list is dependent on a variety of factors including:
+The <a data-cite="?VC-DATA-MODEL-2.0#validity-period">validity period</a> that
+an issuer might choose to express in a status list is dependent on a variety of
+factors including:
       </p>
       <ul>
         <li>
-How often do status values change? In real-time? Daily? Weekly? Monthly?
+How often do status values change? In real-time? Daily? Weekly? Monthly? Other?
         </li>
         <li>
 Is there a regulatory requirement that compels an [=issuer=] to notify a
@@ -1698,23 +1699,23 @@ Is there a period of time after which an [=issuer=] would incur reputational
 damage by not notifying a [=verifier=] of a status change?
         </li>
         <li>
-Is there an expectation by a [=verifier=] to not accept a
-[=verifiable credential=] with a status list that does not expire for an
-excessive period of time?
+Is there any expectation that a [=verifier=] will not accept a
+[=verifiable credential=] with a status list that does not expire for a
+period of time it deems excessive?
         </li>
         <li>
-Is there an excessive network bandwidth or computing burden placed upon an
-[=issuer=] or a [=verifier=] if a validity period is too short for a
-status list?
+Will a short validity period for a status list cause a significant network
+bandwidth or computing burden for an [=issuer=] or a [=verifier=]? Might
+this burden be mitigated by a longer validity period?
         </li>
       </ul>
 
       <p>
-Since these factors vary based on ecosystem and credential type, there is no
-minimum or maximum validity period that is suggested for a status list.
-[=Issuers=] will need to take various considerations into account that are
-specific to their [=verifiable credential=] types and pick validity periods that
-will strike the right balance in their ecosystem.
+Since these factors vary with the ecosystem and credential type, there is no
+minimum or maximum validity period that is suggested for all status lists.
+[=Issuers=] will need to consider various factors that are specific to their
+[=verifiable credential=] types and choose validity periods that will strike
+the right balance in their ecosystem.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -1687,20 +1687,25 @@ status list is dependent on a variety of factors including:
       </p>
       <ul>
         <li>
+How often do status values change? In real-time? Daily? Weekly? Monthly?
+        </li>
+        <li>
 Is there a regulatory requirement that compels an [=issuer=] to notify a
-[=verifier=] that the status of a [=verifiable credential's=] has changed?
+[=verifier=] that the status of a [=verifiable credential=] has changed?
         </li>
         <li>
 Is there a period of time after which an [=issuer=] would incur reputational
 damage by not notifying a [=verifier=] of a status change?
         </li>
         <li>
+Is there an expectation by a [=verifier=] to not accept a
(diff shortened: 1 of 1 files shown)


David-Chadwick (approved):
Given that the status list is a bit string, then the overall size should be small. So I do not think that excessive bandwidth should be a serious impediment to a frequent refreshing of it. Suggest removing the last list item

msporny:
> Given that the status list is a bit string, then the overall size should be small. So I do not think that excessive bandwidth should be a serious impediment to a frequent refreshing of it. Suggest removing the last list item

Remember that the size of the bit string is only one of the concerns. The other concerns is the size of the verifier community, because that's where the real multiplier comes in. For example, if you have a bitstring that is 125K in size, and a verifier population that is 100, then you're looking at 12MB per refresh cycle (which seems reasonable). If, however, you have a verifier population that is 10,000, then you're looking at 1,221 MB per refresh cycle (which is significantly higher).

David-Chadwick:
I take your point. But you are assuming that every verifier downloads every list update every time. But this depends upon the frequency of user interaction at the verifier. If the user interaction is less than the update frequency then it will not be true. So a nightclub that only opens at weekends with an issuer that updates daily, only 2/7 updates will be retrieved by this verifier. OTOH an online shop that has thousands of customers daily, then downloading the list daily is negligible overhead for this verifier. Can we assume that the verifiers will be in different time zones if there are 10,000 of them, so they wont all be downloading at the same time? I think your CA DMV trial should be able to get some really good metrics for use of the mDL at thousands of local shop locations, with the frequency that driving licenses are revoked by CA forcing the list to be updated.

dlongley:
> I take your point. But you are assuming that every verifier downloads every list update every time. But this depends upon the frequency of user interaction at the verifier. If the user interaction is less than the update frequency then it will not be true. So a nightclub that only opens at weekends with an issuer that updates daily, only 2/7 updates will be retrieved by this verifier. OTOH an online shop that has thousands of customers daily, then downloading the list daily is negligible overhead for this verifier. Can we assume that the verifiers will be in different time zones if there are 10,000 of them, so they wont all be downloading at the same time?

That these all seem like important considerations for the issuer indicates to me that we should leave the bandwidth language in place.

It's also worth mentioning that a bitstring with many randomly flipped bits will not compress well -- so this should be avoided if possible or at least taken into consideration with respect to bandwidth as well.

index.html
```diff
@@ -1678,6 +1678,45 @@ <h3>Bitstring Encoding</h3>
       </p>
     </section>
 
+    <section class="informative">
+      <h3>Validity Periods</h3>
+
+      <p>
+The <a data-cite="?VC-DATA-MODEL-2.0#validity-period">validity period</a> of a
+status list is dependent on a variety of factors including:
```
dlongley, line 1686: One reading of this is that the validity period of the status list can't just be read from the status list VC, which isn't true. I think this section is about how issuers might *choose* a validity period, not how it is expressed, definitively, in the status list VC. So I recommend a slight tweak here:

```suggestion
The <a data-cite="?VC-DATA-MODEL-2.0#validity-period">validity period</a> that
an issuer might choose to express in a status list is dependent on a variety of
factors including:
```

dlongley (changes requested):

index.html
```diff
@@ -1678,6 +1678,45 @@ <h3>Bitstring Encoding</h3>
...
+      </p>
+      <ul>
+        <li>
+How often do status values change? In real-time? Daily? Weekly? Monthly?
```
TallTed, line 1690:
```suggestion
How often do status values change? In real-time? Daily? Weekly? Monthly? Other?
```

index.html
```diff
@@ -1678,6 +1678,45 @@ <h3>Bitstring Encoding</h3>
...
+[=verifier=] that the status of a [=verifiable credential=] has changed?
+        </li>
+        <li>
+Is there a period of time after which an [=issuer=] would incur reputational
+damage by not notifying a [=verifier=] of a status change?
+        </li>
+        <li>
+Is there an expectation by a [=verifier=] to not accept a
+[=verifiable credential=] with a status list that does not expire for an
+excessive period of time?
```
TallTed, line 1703:
```suggestion
Is there any expectation that a [=verifier=] will not accept a
[=verifiable credential=] with a status list that does not expire for a
period of time it deems excessive?
```

index.html
```diff
@@ -1678,6 +1678,45 @@ <h3>Bitstring Encoding</h3>
...
+        </li>
+        <li>
+Is there an excessive network bandwidth or computing burden placed upon an
+[=issuer=] or a [=verifier=] if a validity period is too short for a
+status list?
```
TallTed, line 1708:
```suggestion
Will a short validity period for a status list cause a significant network
bandwidth or computing burden for an [=issuer=] or a [=verifier=]? Might
this burden be mitigated by a longer validity period?
```

index.html
```diff
@@ -1678,6 +1678,45 @@ <h3>Bitstring Encoding</h3>
...
+        </li>
+      </ul>
+
+      <p>
+Since these factors vary based on ecosystem and credential type, there is no
+minimum or maximum validity period that is suggested for a status list.
+[=Issuers=] will need to take various considerations into account that are
+specific to their [=verifiable credential=] types and pick validity periods that
+will strike the right balance in their ecosystem.
```
TallTed, line 1717:
```suggestion
Since these factors vary with the ecosystem and credential type, there is no
minimum or maximum validity period that is suggested for all status lists.
[=Issuers=] will need to consider various factors that are specific to their
[=verifiable credential=] types and choose validity periods that will strike
the right balance in their ecosystem.
```

TallTed (changes requested):
Editorial...

dlongley (approved):

msporny:

PR #192, opened 2024-12-02
Added the context references to the vocabulary.yml file

iherman:
This means that the references to the relevant context files will appear for each term. (See the [di vocabulary description](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html#context-files) as a reference to what this means).

### DIFF
--- a/vocab/vocabulary.yml
+++ b/vocab/vocabulary.yml
@@ -1,7 +1,7 @@
 vocab:
   - id: cs
     value: https://www.w3.org/ns/credentials/status#
-    context: https://www.w3.org/ns/credentials/v2
+    context: https://www.w3.org/ns/credentials/status/v1
 
 prefix:
   - id: cred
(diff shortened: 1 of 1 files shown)


msporny (approved):
LGTM.

@iherman, as an aside, I had to remove the context definition because it was breaking the build. The JSON Schema was failing because of the context declarations, but only when using the versions of node used by the Github Action (Node 18, IIRC). In any case, just letting you know since it breaks the spec build (by failing to generate the vocabulary.(html|ttl|jsonld) files).

iherman:
> @iherman, as an aside, I had to remove the context definition because it was breaking the build. The JSON Schema was failing because of the context declarations, but only when using the versions of node used by the Github Action (Node 18, IIRC).

I do not understand that. I use, on my local machine, note version 22.7; the DI repository, in its gh-pages action uses version 16, I do not see any reason why it would not work on version 18.

The version of yml2vocab is the same on my machine as on npm. As for the schema, I have checked the schema file itself on the repository: [Line 164](https://github.com/w3c/yml2vocab/blob/a8ede7cf7dade6214e5c8a7163ce6899326c5263/dist/lib/vocab.schema.json#L164) refers to the possibility of a context, referring to a schema definition at [line 329](https://github.com/w3c/yml2vocab/blob/a8ede7cf7dade6214e5c8a7163ce6899326c5263/dist/lib/vocab.schema.json#L329).

It looks as if the action did not copy the right JSON schema file. No idea why, but GitHub actions are always a mystery to me...

dlongley (approved):

msporny (approved):

msporny:
Editorial, multiple reviews, no changes requested, no objections, merging.

PR #199, opened 2025-02-23
Remove status message feature due to lack of implementations.

msporny:
This PR removes the status message feature from the specification due to a lack of implementations for the feature:

https://w3c.github.io/vc-bitstring-status-list-test-suite/#Data%20Model:%20BitstringStatusList%20Entry

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/pull/199.html" title="Last updated on Feb 23, 2025, 6:04 PM UTC (f574921)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-bitstring-status-list/199/4457728...f574921.html" title="Last updated on Feb 23, 2025, 6:04 PM UTC (f574921)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -560,13 +560,6 @@ Used to temporarily prevent the acceptance of a [=verifiable credential=].
 This status is reversible.
                       </td>
                     </tr>
-                    <tr>
-                      <td>`message`</td>
-                      <td>
-Used to convey an arbitrary message related to the status of the
-[=verifiable credential=].
-                      </td>
-                    </tr>
                   </tbody>
                 </table>
               </td>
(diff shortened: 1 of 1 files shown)


dlongley (approved):

David-Chadwick (commented):
I think it's a shame to remove this feature, as I can see the utility of it. If no-one has implemented it yet, I suggest this is due to minimalist implementations that want to reduce the cost of implementation, rather than the utility of the feature. Once resources are more abundant I suggest that people will want this feature.
Therefore cannot we make it optional rather than mandatory, and then it does not have to be implemented immediately.

msporny:
> If no-one has implemented it yet, I suggest this is due to minimalist implementations that want to reduce the cost of implementation, rather than the utility of the feature. Once resources are more abundant I suggest that people will want this feature. Therefore cannot we make it optional rather than mandatory, and then it does not have to be implemented immediately.

It was always optional, people just chose to not implement it and now we are out of time. We've given the feature over a year to gather implementations and it failed to do so. A last-minute effort to save the feature is not what we need right now. If people want to implement the feature, there are static CR drafts that include the feature that can be cited as well as a test suite that demonstrates that the feature has been implemented correctly. People can try again in v2.0 if they'd like to add this feature, nothing we're doing here prevents organizations from implementing the feature per the 2nd CR snapshot, which was stable.

I will note that a number of the implementers, ourselves included, think this is a dangerous feature and actively argue against using it with our customers. There are serious privacy issues related with it, both from a individual privacy, and a corporate privacy perspective, that we've found very difficult to implement correctly.

All that to say, this is the path the WG suggested the Editor's take on the call last week and I'm implementing what we had consensus for during the last WG call:

https://www.w3.org/2017/vc/WG/Meetings/Minutes/2025-02-19-vcwg#section1-5

mkhraisha:
We have an implementation for this feature, so does @brentzundel at TradeVerifyd https://github.com/w3c/vc-bitstring-status-list-test-suite/issues/61

We are passing all tests locally (by bypassing jsonld safe mode) because of missing context terms in the BitstringStatusListEntry we are unable to pass all the test suite test on the current live version.

I have pushed a PR that fixes the test fixtures https://github.com/w3c/vc-bitstring-status-list-test-suite/pull/63 to fix the tests for the StatusMessage feature, however we will not pass it until the context has been fixed.

I have opened an issue under bitstringstatus list ( not sure if that or the Data model repo is the more appropriate one) to outline the missing context terms: 
https://github.com/w3c/vc-bitstring-status-list/issues/200

msporny:
Based on https://github.com/w3c/vc-bitstring-status-list/pull/199#issuecomment-2679615210, we have two independent implementations now, closing this PR as unnecessary.

iherman:
> Based on [#199 (comment)](https://github.com/w3c/vc-bitstring-status-list/pull/199#issuecomment-2679615210), we have two independent implementations now, closing this PR as unnecessary.

Yay! I trust they will appear in the implemenation report soon.

PR #215, opened 2026-09-19
Updated respecConfig for former editors

iherman:
The pubrules, by default, require that all editors are WG members. However, it is possible, by adding a `<span class="former">…</span>` to the editors' name, to bypass that rule and keep the persons' name on the list. Also, I was helped by @deniak to find a way to do that by slighty modifying the structure of the persons' data in `respecConfig`.

This PR makes those changes for all editors, marking Mike and Orie as "former". Although only these two required special treatment, I did it for all for consistency's sake.


wes-smith:
Under this model, Dave Longley should be re-added as an editor of just v1.0.

wes-smith:
@KDean-Dolphin should also be added as an editor for v1.1.

iherman:
I have added Dave and Kevin to the editors' list in https://github.com/w3c/vc-bitstring-status-list/pull/215/commits/89ab4a2a84b3980dcbf64bc2673f4eeeb5a19c38

index.html
```diff
@@ -62,28 +62,42 @@
...
           w3cid: 152091,
-          note: "v1.1"
+          extras: [{ name: "(v1.1)"  }]
         }, {
+          name: "Dave Longley",
+          url: "https://github.com/dlongley",
+          company: "Digital Bazaar",
+          companyURL: "https://www.digitalbazaar.com/",
+          w3cid: 48025,
+          extras: [{ name: "(v1.0 - v1.1)" }]
```
dlongley: I'm not a current editor, was just a former editor for v1.0:

```suggestion
          extras: [{ name: "(v1.0)", class: "former" }]
```

iherman: To make it clear: adding a "former" means that the system does not list you as current editor. It does not affect the visual output on the page. Which is ok, I believe that is your intention, but I just want to be sure.

